PR #39, opened 2026-01-31
HEPH-31: Creating Heph Language Server

Beloin:
There's a lot to unpack here.

1. Why not use [starlark-lsp](https://github.com/tilt-dev/starlark-lsp)?
- Too weird. It doesnt have any api to ease or way into custom capabilities
- Also their way to import code was so weird, a custom python file had to be generated to be imported into entrypoint.

2. Why not Use Starlark Parser?
- We won't have custom queries.
- Parsers generate AST not an CST.
- In CST I can look direct into nodes an query specified node in position offset (thus querying and matching position).

3. Why use Python Parser?
- Closer to starlark, we don't have a official starlark parser yet.

4. How to test?
- I have also implemented a Neovim plugin that uses this LSP.
- Available [here](https://github.com/Beloin/heph.nvim) .

## Initial Implemented LSP Capabilities

- Completion
- Declaration
- Hover
- References
- Sync

## Not implemented (yet)

- Heph/Starlark modules and structs
- Meaning we dont have any capabilites for `heph.canonicalize` etc.

## Known improvements

1. Use DAG to save relationships from BUILD files
2. Use trie or other prefix trees to make in-memory queries faster

## Examples

Function Doc:
image

Arg Doc:
image

Completions:
image


cmd/heph/utils.go
```diff
@@ -11,10 +11,7 @@ func ValidArgsFunctionTargets(cmd *cobra.Command, args []string, toComplete stri
 	}
 
 	directive := cobra.ShellCompDirectiveNoFileComp
-	isFuzzy, suggestions := autocompleteTargetName(targets, toComplete)
-	if isFuzzy {
-		directive |= cobra.ShellCompDirectiveNoMatching
-	}
```
Beloin: v1.8.0+ removes ShellCompDirectiveNoMatching.

PR #43, opened 2026-04-19
fix PackagePrefix to match on path-component boundaries (HEP-9)

raphaelvigee:
## Summary

- Adds `Matcher::match_addr` method to `htmatcher`
- `PackagePrefix` now requires the package to equal the prefix or start with `{prefix}/`, preventing partial component matches (e.g. `foo/ba` no longer matches `foo/bar/baz`)
- Empty prefix continues to match all packages

## Test plan

- [ ] Unit tests in `src/htmatcher/matcher.rs` cover exact match, child match, partial-component non-match, and empty prefix

Fixes [HEP-9](https://linear.app/heph/issue/HEP-9)

### DIFF
--- a/src/htmatcher/matcher.rs
+++ b/src/htmatcher/matcher.rs
@@ -10,3 +10,54 @@ pub enum Matcher {
     And(Vec<Matcher>),
     Not(Box<Matcher>),
 }
+
+impl Matcher {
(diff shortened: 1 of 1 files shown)


raphaelvigee:
Closing — wrong approach. Will reimplement with a proper Pkg type.

PR #63, opened 2026-06-06
Configurable fs skip via htwalk::Ignore (dirs + globs), engine-owned

raphaelvigee:
## Summary

Replaces the fs plugin's hardcoded skip lists with a single, engine-owned
ignore model shared by every tree-walking plugin, and makes it configurable.

### Config — `fs.skip`
A top-level block in `.hephconfig2`, all entries **root-relative**:

```yaml
fs:
  skip:
    - node_modules          # literal dir → prune (also ./node_modules, node_modules/)
    - foo/**/bar            # glob, literal tail → prune every bar dir under foo
    - some/*.go             # glob, wildcard tail → file exclusion only (no prune)
```

`.git` is an always-on built-in (like the heph home), never hardcoded in pluginfs.

### `src/htwalk` — shared `Ignore`
New module used by the fs driver/provider and the buildfile/go providers
(replaces three duplicate `SkipMatcher`/`FsSkip` types). It distinguishes
**pruning** a directory subtree from **excluding** files, deriving the
directory matcher from the glob shape — no sentinel-child probe:
- glob whose final component is a literal names directories → prune (`foo/**/bar`);
- recursive `…/**` / `…/**/*` → prune via its prefix (`**/node_modules/**` → `**/node_modules`);
- wildcard-tail glob (`some/*.go`) → file-only;
- exact `dirs` (heph home + literal `fs.skip`) → prune by path (component-wise, so `node_modules/` works).

### Engine plumbing
- `PluginInit { root, skip_dirs, skip_globs }` injected into every register method.
  `register_provider`/`register_driver`/`register_managed_driver` take a closure
  receiving it; `*_factory` variants compose with `&Options`.
- Fallible `try_register_provider` / `try_register_driver`: the built-in `fs`
  plugin composes its `Ignore` from `PluginInit` inside the closure, so a bad
  `fs.skip` glob surfaces as an error from `Engine::new`.
- `fs` is an always-on built-in registered by `Engine::new` (like plugingroup/pluginquery).

### list & get honor the ignore
Package discovery already pruned skipped dirs; now buildfile/go `list` and `get`
also return empty / NotFound for a package inside a pruned subtree
(`Ignore::prunes_package` walks the package dir + ancestors). So a direct
`heph run //node_modules/x:y` no longer resolves. (stdlib/third-party go
packages live outside the workspace, so skips don't apply.)

## Tests
- `htwalk`: prune-vs-exclude, literal-tail / recursive / wildcard-tail globs,
  `./` normalize, trailing slash, ancestor `prunes_package`.
- config_file / bootstrap: `fs.skip` split into dirs+globs, `./` normalize, `.git` built-in.
- pluginfs / buildfile: skip-dir & skip-glob pruning in walks; `get`/`list` skip pruned packages.
- testkit: e2e wires fs via `Engine::new` auto-registration.

`cargo clippy --all-targets --locked -- -D warnings` clean; 987 lib + plugingo-e2e green.

### DIFF
--- a/src/pluginfs/mod.rs
+++ b/src/pluginfs/mod.rs
@@ -109,16 +109,29 @@ impl EProvider for Provider {
 
     fn list_packages<'a>(
         &'a self,
-        _req: ListPackagesRequest,
+        req: ListPackagesRequest,
         _ctoken: &'a (dyn Cancellable + Send + Sync),
     ) -> BoxFuture<
         'a,
         anyhow::Result<Box<dyn Iterator<Item = anyhow::Result<ListPackageResponse>> + Send>>,
     > {
         Box::pin(async move {
-            let items: Vec<anyhow::Result<ListPackageResponse>> = vec![Ok(ListPackageResponse {
-                pkg: PkgBuf::from(PKG),
-            })];
+            // `fs` owns exactly one synthetic package (`@heph/fs`). Yield it only
+            // when it falls under the requested prefix, so a narrowed query (e.g.
+            // `//a/b/c:...`) doesn't spuriously surface it.
+            let prefix = req.prefix.as_str();
+            let under_prefix = prefix.is_empty()
+                || prefix == PKG
+                || PKG
+                    .strip_prefix(prefix)
+                    .is_some_and(|rest| rest.starts_with('/'));
+            let items: Vec<anyhow::Result<ListPackageResponse>> = if under_prefix {
+                vec![Ok(ListPackageResponse {
+                    pkg: PkgBuf::from(PKG),
+                })]
+            } else {
+                vec![]
+            };
             Ok(Box::new(items.into_iter())
                 as Box<
                     dyn Iterator<Item = anyhow::Result<ListPackageResponse>> + Send,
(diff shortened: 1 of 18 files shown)


src/pluginfs/mod.rs
```diff
@@ -348,14 +362,79 @@ impl ProviderFn for BaseFn {
...
+    /// matcher for the common case of a target with no user excludes.
+    builtin_any: Arc<wax::Any<'static>>,
+}
+
+impl FsSkip {
+    /// Builds the skip set from the engine-provided `skip_dirs` (absolute paths).
+    /// Each skip dir located under `root` also contributes a `<rel>/**` exclude
+    /// glob so the compiled `not` matcher agrees with the walk-time pruning.
+    pub fn new(root: &std::path::Path, skip_dirs: &[std::path::PathBuf]) -> anyhow::Result<Self> {
+        let mut exclude_globs = vec!["**/.git/**".to_string()];
```
raphaelvigee: the .git folder should be passed in by the engine

raphaelvigee: this needs removal from here, .git should be ignored at the engine level as a dir

PR #81, opened 2026-06-12
feat(buildfile): BUILD-file LSP

raphaelvigee:
## What

A Language Server for BUILD files, launchable two ways:
- the hidden command `heph tool build-lsp`, and
- a synthetic, **uncached** target `//@heph/build:lsp` (exposed by the `pluginbuildfile` provider, `exec` driver, re-execs the current heph binary into `tool build-lsp`).

Built on the official **`starlark_lsp`** crate via an `LspContext` impl, plus a thin stdio↔in-memory **proxy** that enriches `hover`/`completion` responses the stock server can't produce.

## Features

- **Builtin symbol resolution** — `target` / `file` / `glob` / `struct` / `provider_state`, `heph.core.*`, and provider functions complete and hover, sourced from the same globals BUILD evaluation uses (now carrying doc comments).
- **Target-address autocomplete** — `//pkg:…` completes packages, `//pkg:name` completes target names.
- **Goto-definition** on address strings — both absolute `//pkg:name` and relative `:name`.
- **Driver schema** — `Driver`/`ManagedDriver` gain an optional `schema()` (implemented for `exec`); completion inside a `target(driver="exec", …)` call offers that driver's config fields with types/docs, via `Engine::driver_schema`.
- **Provenance hover** — every target records its source call-stack provenance, so hovering a `target()` call (or a macro that produced several targets) shows a copyable **"Generated targets"** block. Capture is gated to the LSP eval path, so normal builds pay nothing.

## How it fits together

```
editor ──stdio──▶ heph tool build-lsp ──▶ proxy ──▶ starlark_lsp(HephLspContext)
                                           └─ enriches hover (provenance) + completion (driver schema)
```

The proxy is required because `LspContext` has no hover hook and can't do call-context-dependent completion. Everything else (diagnostics, builtin completion/hover, load resolution, address completion/goto) is pure `LspContext`.

## Verification

- New unit tests: exec `schema()`, provenance capture incl. macro call-site chains, `DocIndex` hover/driver lookup, synthetic-target `get`/`list`/`list_packages`, provenance-off on the normal eval path.
- End-to-end smoke against the `example/` workspace through a real engine: builtin hover shows docs; hovering a macro call returns the generated-targets block; completion inside `target(driver="exec")` includes the exec fields; `inspect spec //@heph/build:lsp` resolves to the exec spec with cache off.
- Full `cargo test` green; `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean.

## Notes / follow-ups

- BUILD-file patterns default to `BUILD` in the LSP (not yet read from workspace config).
- Engine bootstrap logging should stay on stderr (stdout is the JSON-RPC channel); verified in the smoke test.

### DIFF
--- a/.github/workflows/heph.yml
+++ b/.github/workflows/heph.yml
@@ -31,6 +31,18 @@ jobs:
       - uses: cachix/install-nix-action@v31
         with:
           extra_nix_config: fallback = true
+      # Persist the Nix store (rust-overlay toolchain, devenv env, …) across runs.
+      # Cold runners otherwise re-realize the whole toolchain every job, which
+      # intermittently fails to fetch/realize a rust-overlay patch
+      # ("path '/nix/store/…fix-kill-doctest.patch' is not valid"), forcing reruns.
+      # Keyed on the devenv lock so a toolchain bump busts the cache.
+      - name: Cache Nix store
+        uses: nix-community/cache-nix-action@v6
+        with:
+          primary-key: nix-${{ runner.os }}-${{ hashFiles('devenv.lock', 'devenv.nix', 'devenv.yaml') }}
+          restore-prefixes-first-match: nix-${{ runner.os }}-
+          gc-max-store-size-linux: 5368709120
+          gc-max-store-size-macos: 5368709120
       - name: Install devenv.sh
         run: nix profile add nixpkgs#devenv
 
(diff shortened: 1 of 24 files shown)


raphaelvigee:
Follow-up: LSP no longer hardcodes `BUILD` — it now reads the buildfile provider's `patterns` from workspace config (commit 6fce0787). Verified with a `*.BUILD` workspace: completion discovers a package whose only file is `b.BUILD`, and goto-def resolves into it.

PR #135, opened 2026-06-27
feat(plugin-go): per-package go vet/lint with serialized facts

raphaelvigee:
## What

Per-package Go `lint` target with **nogo-style serialized facts** — interprocedural analyzers (printf wrappers, lostcancel, …) keep full fidelity across first-party package boundaries, while caching per package. Linter selection + suppression via **`.golangci.yml`**.

## Why not golangci-lint directly

golangci-lint loads the whole graph in one process and keeps go/analysis facts in memory only — no fact-file export, incompatible with a per-package build cache. It also hides its analyzers (`goanalysis.Linter.getAnalyzers` is unexported) and its runner doesn't serialize facts, so it can't be reused as a library here.

But golangci **doesn't own the analyzers** — it curates upstream `go/analysis` packages that export theirs.

## Approach

- Run **`x/tools` unitchecker** (the per-unit driver `go vet` uses): dep type info via `PackageFile`, dep facts via `PackageVetx`, own facts out via `VetxOutput`.
- **`heph-govet`** imports the same upstream analyzers golangci curates — `go vet` passes + honnef `staticcheck`/`gosimple`/`stylecheck` — and parses `.golangci.yml` itself (`linters.default/enable/disable`, `linters.exclusions.rules`) to select + suppress. No dependency on golangci-lint.
- heph's content-hash caching makes a package re-lint only when its own sources, a dep's archive, a dep's facts, the analyzer binary, or `.golangci.yml` change.

## Targets

- **`_lint`** (`go_lint`) — runs heph-govet, writes `lint.facts` (consumed by dependents) + `lint-report.json`. **Exits 0 even with findings** so facts always cache.
- **`lint`** (`go_lint_gate`) — consumes `_lint`'s report, **fails on findings**.
- **`lint-fix`** (`go_lint_fix`) — applies the report's suggested fixes back into source via **`codegen=in_place`**.
- **`format`** (`go_format_check`) — fails on any unformatted file (writes nothing).
- **`format-fix`** (`go_format`) — reformats sources and writes back via **`codegen=in_place`**.

Facts flow **first-party-only** (bounds the graph); std/thirdparty give type info via their `build_lib` archives.

`.golangci.yml` at the workspace root is a hashed `config` input, passed via `HEPH_GOVET_GOLANGCI_CONFIG`. Absent → the standard set.

## `.golangci.yml` coverage

- **linter selection** — `linters.default/enable/disable`.
- **`linters.settings`** — govet (`enable`/`disable`/`enable-all`/`disable-all` + per-analyzer flags like `printf.funcs`, `shadow.strict`); honnef `staticcheck`/`gosimple`/`stylecheck` (`checks` patterns; `initialisms`, `dot-import-whitelist`, `http-status-code-whitelist`). govet flags are set via `Analyzer.Flags.Set` before `unitchecker.Main` (they survive its flag parse — our argv carries none); honnef knobs override `config.Analyzer.Run` in place; `checks` filters the analyzer set.
- **suppression** — `//nolint[:l]`, `linters.exclusions.rules` (linter/path/text).
- **`formatters`** — `enable` + `settings` for gofmt/gofumpt/goimports.

## Suppression (honored at analysis time)

- **`//nolint[:l1,l2]`** — trailing covers its line; own-line also covers the next line; bare covers all. Matched against the golangci linter name (`//nolint:gosimple` suppresses `S1002`).
- **`linters.exclusions.rules`** — per-rule linter scoping + path/text regexes.

Implemented by wrapping each analyzer's `Pass.Report`, so the report is already filtered (the gate needs no change).

## Fix (`lint-fix`)

`lint-fix` (`go_lint_fix`) rewrites a package's Go sources with go/analysis suggested fixes and writes them back over the tracked files via **`codegen=in_place`**.

- **Reuses `_lint`'s `-json` report** — which already carries each diagnostic's `suggested_fixes` (byte-offset text edits) — so it runs **no analysis of its own** and shares `_lint`'s per-package cache. Inputs: that report + the package's own `.go` sources. Outputs: the rewritten sources, declared `CodegenMode::InPlace`, so the engine writes them back on a top-level run (and reaches a fixpoint on re-run).
- Application is purely mechanical, in two pure helpers: `report_edits` (parse → edits grouped by source basename; **first** suggested fix per diagnostic, the go/analysis auto-fix convention) and `apply_edits` (sort by offset, drop out-of-range/overlapping keeping the earliest, splice in one pass). Offsets line up because `_lint` and `lint-fix` stage byte-identical sources.
- Sources stage as **writable copies** (the default, non-read-only stage path), so the in-place overwrite can never touch a shared cache blob.
- `.golangci.yml` selection flows in through `_lint`, so the fix target needs no config wiring.
- No Go / heph-govet changes — the report already emits `suggested_fixes`.

## Formatters (`format` / `format-fix`)

heph-govet doubles as the formatter driver in a **`-format`** mode (a whole-file rewrite path, separate from unitchecker), so formatting reuses the one hermetic Go binary — no second ~50-min tool build. Backends: **gofmt** (`go/format` + golangci/gofmt for `-s` + rewrite rules), **gofumpt**, **goimports** (`FormatOnly` — sort/group/remove; *adding* imports needs the module graph and would break the hermetic sandbox, so it is not attempted). `format-fix` mirrors `lint-fix` (`codegen=in_place`); `format` is the check gate.

## Not supported in the per-package model (documented)

- `unused` (and `unparam`) — whole-program. Feasible via the facts model + a root fan-in target (honnef's `SerializedGraph.Merge` is exactly that combine); designed, not yet built.
- goimports import-*adding* (needs the module graph); exclusion presets + `rules[].source` (only path/text/linter parsed).

## Tests

- Rust: spec builders, report parser, gate wiring, config-group wiring; provider resolution (real `go list`).
- Go: `scanNolint` + `suppressed` unit tests (nolint scoping, exclude text/path/linter).
- Smoke: gosimple **S1002** + stylecheck **ST1000** fire under unitchecker `-json` exit 0; 95 staticcheck analyzers wired; unknown linter fails; `//nolint:gosimple` and a text exclude-rule drop S1002 while ST1000 survives; mis-scoped rule does not suppress.
- `lint-fix`: `report_edits` / `apply_edits` (overlap, insertion, out-of-range, first-fix-only), spec builder, an async `parse()` check asserting in-place pkg-relative outputs, and provider resolution over real `go list`.
- settings: govet resolve (enable/disable/all) + flag application (printf.funcs, unknown-name failure), honnef check filtering (all/none/`-`/glob), honnef config override (initialisms + default fallback).
- formatters: each backend (reformat, `-s` simplify, gofumpt extra rules, goimports sort, idempotence), plus Rust spec builders + provider resolution.
- prod clippy clean; `go test ./...` green.

## Distributing `heph-govet`

`heph-govet` is **heph's own** tool, not the user's: a workspace consuming the go plugin has no `tools/heph-govet` package to build it from, so resolving it as `//tools/heph-govet:build` only ever worked inside this repo. It now ships as a release asset, and the plugin fetches it through a new built-in driver.

### `http_fetch` (new built-in driver, `crates/plugin-http`)

Downloading a file off the internet isn't a go-plugin concern, so it's a driver of its own — always-on, like `nix` / `textfile`, and usable from any BUILD file:

```python
target(name = "dl", driver = "http_fetch", url = "https://…/tool_{goos}", sha256 = "…", executable = True)
```

- **The URL templates over the target's addr args** — `…/heph-govet_{goos}_{goarch}` on `:heph-govet@goos=darwin,goarch=arm64` — so one definition serves every platform.
- The **rendered** URL feeds the target hash, so each arg combination is its own cache entry, and an unresolvable placeholder is an error rather than a silently mis-fetched file.
- Optional `sha256` verified before the bytes are written (mismatch fails the build closed); `executable` marks the result runnable. One fetched file = one cacheable output.

### CI publishes the binaries

The `Build heph-govet` job cross-compiles it for the platforms heph publishes (linux amd64/arm64, darwin arm64) — pure Go, one runner, ~1 min — and publishes `heph-govet_<os>_<arch>` as release assets alongside the CLI and the plugin cdylibs. Their SHA-256s are job outputs the Rust `build` job **bakes into the plugin** (`HEPH_GOVET_SHA256_<OS>_<ARCH>`), which is why `build` now waits on it: nothing for a user to configure, and a tampered asset fails closed.

### `govet` provider option — an addr

- **unset** → `//@heph/go/govet/<tag>:heph-govet`, the `http_fetch` target the provider synthesizes over the asset published in the release this plugin was built from (the two always agree — same CI run).
- **a build target** (`govet = "//tools/heph-govet:build"`) → run a heph-govet built from source. What heph's own repo, its tests, and dev builds use: `v0.0.0-dev` has no release to download from (a dev build that forgets to set it gets that as an error, not a 404 mid-build).
- **another release's download target** → pin it, with a `checksums` entry (key `govet/<tag>/<os>/<arch>`) if you want it verified.

An addr given without args gets the *host's* factors: the tool always runs natively, whatever platform the analyzed code targets.

Tests: `http_fetch` unit tests (template rendering incl. failure modes, checksum, exec bit, per-arg hashing) plus e2e against a loopback server — fetched tool staged into a bash consumer's sandbox and executed, checksum mismatch fails the build, each templated arg fetches once and then hits cache.

## Remaining

The end-to-end `ws.run("//:lint")` / `ws.run("//:lint-fix")` tests still need a released `heph-govet` to download (this PR's CI publishes the first one), or `govet = "source"`, which exercises the in-sandbox thirdparty download of x/tools + honnef (`go.sum` committed; host build verified).


raphaelvigee:
### In-sandbox build of heph-govet: proven ✅ (with a caveat)

Verified end-to-end that **heph can build `heph-govet` itself** — ran `//:build` on the tool through the engine; the full `x/tools` + `honnef.co/go/tools` thirdparty tree compiles hermetically (each package its own `build_lib`) and links. Test passed.

**Caveat — cold build is ~50 min** (3033s, no heph cache). heph caches per-package so subsequent builds are fast, but **first-run lint latency is brutal**.

**Recommendation for productization:**
- ship `heph-govet` as a **prebuilt/released binary** (what most linter integrations do), or
- rely on a **warm remote cache** so the x/tools+honnef `build_lib`s are pulled, not recompiled.

`govet_addr` is currently hardcoded to `//tools/heph-govet:build`; making it injectable would allow a prebuilt path *and* a fast end-to-end lint test (the only reason the e2e `//:lint` run isn't in CI is the cold build time).

Coverage as-is: unit (spec builders, report parser, gate, config wiring), provider resolution (real `go list`), Go suppressor tests, heph-govet smoke tests — plus this manual proof of the in-sandbox build.

Infra notes for reviewers: needs an isolated `CARGO_TARGET_DIR` (shared target dir links a stale `plugin-go` rlib → phantom "cannot find GoLintDriver"); the parallel e2e suite trips a sqlite "eval cache disk I/O error", single tests are fine.

PR #200, opened 2026-07-27
fix(cache): give the sqlite read pool headroom over the pipe budget

raphaelvigee:
**P2.1** of the threading/scale remediation.

## Problem

`read_pool.max_size` was set to exactly `pipe_limit` — zero headroom. The pipe path holds its pooled connection for the **consumer's whole drain**, not for the SELECT (`reader` hands back a `GuardedReader` over an `io::pipe` while a `rayon::spawn`'d copy holds the connection). A burst of streaming reads the consumer hasn't finished draining therefore owns every connection in the pool for an unbounded time.

Six call sites take a connection with **no pipe permit at all**: `exists`, `list_targets`, `list_target_entries`, `seekable_reader`, and both phases of `reader`. Behind a saturated pool each blocked on r2d2's inherited 30 s default for what is a single indexed lookup — and `exists` sits on the remote-hit path, where it runs inline on a tokio worker.

Two of those are not brief either: `list_targets`' producer thread holds its connection for a whole GC stream, and `seekable_reader`'s `OwnedBlob` holds one for the reader's lifetime.

It also closes a scheduling cycle the audit flagged: a queued `rayon::spawn` pipe-copy owns a connection it cannot release until rayon schedules it, while a rayon worker sits in `read_pool.get()`.

## Change

- `read_pool_size = pipe_limit + read_pool_headroom()`, where headroom is `(4 * cores).clamp(16, 64)` — the unpermitted callers are bounded by the threads that can be inside one at once (tokio workers, the `hcore::blocking` pool, rayon), and the clamp keeps both a 2-core machine and a 128-core one from picking an absurd number.
- `connection_timeout` explicit instead of inherited, and a `read_conn()` helper so a timeout says *the pool was exhausted* rather than naming the query.

The timeout stays at 30 s deliberately. The resize is what removes the queueing; shortening the bound on its own would only convert a shortage that would have cleared into a failed build.

## Test

`undrained_streaming_reads_do_not_starve_an_indexed_lookup` fills the pipe budget with undrained 4 MiB readers and asserts `exists` still answers. Verified it fails with the headroom forced to zero (times out at 5 s against the 30 s pool wait) and passes with it. `with_exact_pipe_limit` exists so the test can saturate 4 pipes instead of the `DEFAULT_MAX_CONCURRENT_PIPES` floor of 64.

## Deliberately not in this PR

Every read connection carries `cache_size = -64000`, so the pool's worst-case page cache scales with its size (this change takes it from `pipe_limit` to `pipe_limit + ≤64`). That ceiling predates this change, and shrinking it is a perf claim that wants measurement — `mmap_size = 256 MiB` means most reads bypass the page cache entirely, but that is an argument, not a number.

Per the plan, `read_pool` is **not** routed through `pipe_sem` — that would close a deadlock cycle through FUSE `copy_up`.

### DIFF
--- a/crates/engine/src/engine/local_cache_sqlite.rs
+++ b/crates/engine/src/engine/local_cache_sqlite.rs
@@ -10,11 +10,50 @@ use std::path::PathBuf;
 use std::sync::mpsc;
 use std::sync::{Arc, Condvar, Mutex};
 use std::thread::JoinHandle;
+use std::time::Duration;
 use tempfile::SpooledTempFile;
 
 const SPOOL_MEM_THRESHOLD: usize = 1024 * 1024;
 pub const DEFAULT_MAX_CONCURRENT_PIPES: usize = 64;
 const WRITE_BATCH_MAX: usize = 64;
+
+/// How long `read_pool.get()` waits before giving up.
+///
+/// With the pool sized above the pipe budget (see [`read_pool_headroom`]) nothing
+/// should ever queue here, so reaching this is a bug — a leaked connection or a scheduling cycle —
+/// and the point of the bound is to say so instead of stalling. It stays
+/// generous because the alternative failure is worse: a pool shortage that would
+/// have cleared turns into a failed build.
+const READ_POOL_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Read connections to keep beyond the pipe budget, for the callers that take one
+/// with **no** pipe permit.
+///
+/// Sizing the pool at exactly the pipe budget leaves zero headroom, and the pipe
+/// path holds its connection for the *consumer's* whole drain — so a burst of
+/// streaming reads can own every connection for an unbounded time while `exists`,
+/// `list_targets`, `list_target_entries`, `seekable_reader` and the SELECT phase
+/// of `reader` are still trying to acquire one. Those then block for
+/// [`READ_POOL_TIMEOUT`] on paths that are supposed to be a single indexed
+/// lookup.
+///
+/// It also breaks a scheduling cycle: a queued `rayon::spawn` pipe-copy owns a
+/// connection it cannot release until rayon runs it, while a rayon worker sits in
+/// `read_pool.get()`. With headroom the unpermitted caller is served from spare
+/// capacity and the copy gets to run.
+///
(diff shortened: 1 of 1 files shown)


raphaelvigee:
Follow-up pushed after `feature-quality` review. Two corrections worth stating explicitly:

**The deferral rationale in the PR body above is wrong and I'm retracting it.** I argued `mmap_size = 256 MiB` means most reads bypass the page cache. `mmap_size` maps the *first* 256 MiB of the file; everything past it is heap-copied into the pcache normally. With a 100 MiB spill threshold this DB routinely exceeds that — so the argument holds precisely where the ceiling is harmless and fails precisely where it isn't. The `cache_size` question is still deferred, but on "this needs a number", not on that reasoning.

What would settle it, for whoever picks it up: `sqlite3_db_status(SQLITE_DBSTATUS_CACHE_USED)` summed across pooled connections at the end of a warm-cache run on a **>256 MiB** cache DB, at pool 64 vs 128 — RSS is the wrong instrument here, since the same physical page faulted through N mappings is counted N times per-PTE on Linux. If the answer is "shrink it", the right knob is a process-wide `SQLITE_CONFIG_PAGECACHE` / `soft_heap_limit64`, not a per-connection cap multiplied by a pool that scales with core count. In-tree control: `crates/walk/src/cached_walker.rs` runs its read pool at 16 with no `cache_size` override at all.

**`pipe_limit` is still unclamped**, so the pool has no absolute ceiling — `--jobs 512` gives `2*512 + 64`. That predates this change (the pool was already `pipe_limit`), and capping it is a decision about what `--jobs` means for memory rather than a mechanical fix, so I've left it and said so in the commit rather than changing `--jobs` semantics in a PR about headroom.

Everything else from the review is in `d12e2b79`: headroom now derives from the caller's `pipe_limit` instead of `available_parallelism()`; the "bounded by the threads that can be inside one" proof language is gone from both docs (it omitted sandboxfuse's per-core session threads, and the clamp falsifies it anyway); `test_on_check_out(false)` pinned; `list_targets` and the exhaustion-error path now tested; the pipe-saturating tests take 2 rather than 4 global rayon workers.

PR #222, opened 2026-07-28
fix(engine): actually enforce cache.history during a run

raphaelvigee:
Item **P6.3** of the concurrency remediation plan (`ai-docs/THREADING_PLAN.md`). Based on `master`, independent of #219 (P6.1), which has since merged.

## `cache.history` has never been enforced by a build

The post-write trim deletes cache revisions, so it needs the target's **write** lock — `gc.rs` documents why a read is not enough, and the lock is per-addr, not per-revision. But it ran inline, on the background lane, the moment `cache_locally` finished, while the request that wrote the revision was still alive.

That request holds a *read* on the same addr. The riding read guard lives in the addr-keyed `mem_locked_result` cell and is cloned into every artifact handed out, and `Memoizer` only evicts on a cycle error — so the read is held for the whole run. `try_write` is non-blocking by design, so it returned `Ok(None)` into a bare empty match arm and skipped. Every time, silently.

Verified against `master` before writing anything, and re-verified against current `master` (which now includes #219): three runs of one target with `cache.history = 1` leave all three revisions on disk, and a following `gc_all` sweep reclaims two of them.

## Behavior restoration — read this part

**This changes observable disk behavior for existing users.** It is not a refactor and not a pure optimization.

A build that rebuilds a target now reclaims that target's revisions beyond `cache.history` **before the process exits**. Previously they accumulated until someone ran `heph gc`. Concretely:

- Cache directories will shrink on the first run after upgrading — possibly a lot, on a repo that has never been swept.
- A workflow that was unknowingly relying on more revisions being present than `cache.history` allows — revert a change, expect an instant cache hit on the older revision — will now see a rebuild instead.

That is the documented meaning of `cache.history`. It simply has not been true until now. Anyone who wants the old effective behavior should raise `cache.history`, which is what it is for.

## The fix

Record `(addr, keep, hashin)` as the revision is written; submit the batch when `RequestStateData` drops. Two things make that work, and both are load-bearing:

**`DeferredTrims` is the last field of `RequestStateData`.** Rust drops a struct's fields in declaration order *after* its own `Drop::drop` returns, so being last is what guarantees every memoizer — and every riding read it holds — is already gone when the trims are submitted. Moving the field earlier reinstates the bug with no other symptom. Called out in doc comments on both the field and the type.

**Dropping the request is still not sufficient on its own** — this was the surprise, and it took a `Backtrace` in `LockedResolution::drop` to find. `hcore::blocking::backstop` registrations are never unregistered: a waiter re-registers on every pending poll, and its last registration sits in the global `PENDING` list until a tick sweeps it, up to 250 ms later. A `Waker` owns its task, and one of those tasks is the `mem_locked_result` cell — so the riding read outlives the request that created it by up to a backstop period, and every trim still found its target contended. `flush_backstop()` hands those registrations back explicitly. Waking early is always sound: a genuinely pending waiter re-registers from the poll it provokes.

The batch takes its `bg_pending` slot when it is **handed to the cleaner**, not when a trim is recorded — see the review section below for why that distinction is the difference between a leak and an unexitable process.

## `bg_pending` gains a contract

Wait on it **only after releasing the request**, because a request's last piece of background work is submitted by its own drop. `heph run` already unwinds that way — the drain loops in `tui::backend` run after the app future, which owns the request, returns — but three in-tree test helpers waited while holding `rs` and are updated to drop it first. Documented on the field.

## Tests

- **`cache_history_is_enforced_by_the_end_of_the_run`** (`crates/e2e`) — three runs, three revisions, `history = 1`, asserting **after every run** that a fresh `gc_all` finds nothing to reclaim. Checking only once at the end would not distinguish "every run enforced its budget" from "only the last one did", since run 3's trim deletes both older revisions anyway.

  Each run uses a fresh engine over the same root (new `Workspace::reopen`), because an in-process re-run replays the memoized spec: the rewritten BUILD file produces the same `hashin` and no second revision. The first draft of this test passed on `master` for exactly that reason and could not have failed — worth knowing, because this shape of test is silently vacuous without the fresh engine.
- `recording_a_trim_takes_no_background_slot`, `no_deferred_trim_means_no_background_work`, `suppressed_request_records_no_trims`, `engine_dropped_first_discards_the_batch` — the counter contract and the teardown corners.
- `flush_backstop_releases_registered_wakers_without_waiting_for_a_tick` and `flush_backstop_does_not_strand_a_still_pending_waiter` — that flushing *drops* registrations rather than merely waking them, and that the re-registration invariant the whole mechanism rests on actually holds.

## Review board

`code-quality` PASS WITH NITS · `hermeticity` HERMETIC · `feature-quality` BLOCKED. All actioned in the second commit.

### The one that mattered: the reservation could have made a failed build unexitable

`code-quality` found it, and it is worse than the race it was protecting against. The batch originally took its `bg_pending` slot when a trim was *recorded* and only gave it back from `DeferredTrims::drop`. But `RequestStateData` can be pinned indefinitely by a reference cycle: an abandoned memoizer cell keeps its incomplete future (`hmemoizer::cell`'s `slot`), that future holds `Arc<RequestState>`, and the memoizer that owns the cell is a field of `RequestStateData`. `fail_fast` fanout drops in-flight awaiters at every fanout site, so this is reachable on ordinary build-failure paths — there is already a regression test in the tree for one instance of it.

Before this PR that leaked memory and an uncancelled token, silently. With the reservation it would have leaked an exit-gating slot, and both TUI drain loops are untimed: `heph run` would never exit after a failed build. SIGKILL only.

Removed. The slot is taken when the batch is handed to the cleaner, exactly as `enqueue` has always worked, and `reserve`/`release_reservation`/`enqueue_reserved` are gone with it — `sandbox_cleaner` no longer exposes an unguarded manual balance protocol whose imbalance is a hang.

`hermeticity` independently argued the *opposite* — that reserve-at-record was necessary, because `spawn_remote_upload`'s `_dec` drops before its captured `rs` and a submit-time reservation exposes a zero window. Both are right about their own half. The window is real, but it is a **drop-order bug in `spawn_remote_upload`**, not something the counter should be contorted to paper over. So it is fixed there: the task now drops its `Arc<RequestState>` before the `Decrement` guard, so the request's trims are submitted while the upload still holds its own slot. No zero window, and no hang.

### Diagnosability

`hermeticity` and `feature-quality` made the same point, and it is the sharpest one in the set: **this fix's failure mode and the bug it fixes were byte-identical in the logs.** `Ok(None) => {}` — a bare empty arm — is exactly how `cache.history` went unenforced for so long without anyone noticing. It now says so, at `debug` (a concurrent build of the same target is ordinary, this fires per addr, and the revision is reclaimed by the next write's trim or by `heph gc`). The engine-already-dropped path was equally silent and now logs too.

### Also fixed

- **`Workspace::reopen` silently dropped `fs_skip`** (`hermeticity`, `code-quality`). `fs_skip` decides which files a tree-walking plugin sees — the declared inputs, hence the `hashin`. A reopened engine that dropped it would write into the same on-disk cache under different keys, and a test asserting a hit across the two would prove nothing. Config is carried explicitly now, and `with_static` workspaces refuse to reopen rather than silently substituting a different provider set.
- **`heph gc` phase 1 raced its own phase 2** (`hermeticity`). Resolution executes shared cacheable deps, each recording a trim that fired when the phase-1 state dropped — concurrently with phase 2's write locks, making `GcStats` report whichever got there first and surfacing a contended-lock notice naming this very process as the holder. The sweep now suppresses its resolution state's trims; it *is* the authoritative trim.
- **A panic in one trim discarded every trim after it** (`code-quality`), since the cleaner's `catch_unwind` wraps the whole job and the label named none of them. Contained per target.
- **`flush_backstop` runs from `Drop`**, where a panicking waker would unwind out of a destructor — and abort if that drop were itself unwinding. Contained. It also returns how many registrations it took, so its test can tell its own flush from a tick that landed first.
- **`hash_only` requests refuse to defer trims.** Unreachable today, asserted so it stays that way.
- `defer_trim` takes `hashin` by value; the drain report names history trims.

### Deliberately not done, with reasons

**Blocking `write` instead of `try_write`** (`feature-quality`). Overruled. The lock is a cross-process `flock`, so a blocking acquire can wait on *another `heph` process* arbitrarily long — on the single FIFO cleaner thread that also owes every sandbox `rm -rf`, on the path that gates process exit. `ResultLock::write` is also `async` and that thread has no runtime. A skip is recoverable (next run, or `heph gc`); an unbounded stall of all cleanup is not.

**The end-of-run serial tail** (both reviewers, MAJOR). Real, and the largest remaining item. The missing within-budget pre-check is exactly what #219 added — with it, a cold run's trims cost one `list_target_entries` each and take no lock at all, which is the dominant term — and sharding the batch across lanes is P6.2's subject. Nothing here is measured for tail latency; that is a `perf-measurement` job on a cold 100k run, timed from summary to exit.

**Making the last-field invariant mechanical** (all three, MINOR). Both alternatives replace it with explicit memoizer teardown in `RequestStateData::drop`. Declining: enumerating "every memoizer that can hold a read guard" is fragile in a quieter way — one added later is silently missed, where last-field is total by construction. The per-run e2e assertion is what catches a violation.

**Draining `bg_pending` outside `heph run`** (`feature-quality`). Real and pre-existing — sandbox rmdirs and remote uploads have always been enqueued on counters no non-`run` command drains. Now that the slot is only taken at submit, the failure mode is "the process may exit before a trim runs", not a hang. Draining in `bootstrap::block_on` would cover every command in one place, but it changes exit behaviour for all of them and deserves its own change.

## Open question — a user decision, not mine

`hermeticity` flags that the trim's success has no formal happens-before edge: `flush_backstop()` wakes, the cleaner's `try_write` follows, and nothing sequences "the woken tasks actually released their read guards" in between.

In the case that actually caused this bug the release *is* synchronous — the registered waker is the `hmemoizer::Cell` itself, and `ArcWake::wake` drops the last `Arc` inline (that is what the backtrace showed). But that is a property of who happens to be awaiting, not a guarantee, and the platform where dropped wakes are a documented hazard is macOS — which is precisely why `hcore::blocking`'s backstop exists at all.

So the honest position: this may be best-effort rather than guaranteed, and if it degrades it degrades on one OS. That is an OS-axis behaviour split, and per `CLAUDE.md` it is yours to settle, not mine.

- **A — accept best-effort.** Zero cost. A lost race is now visible in `debug` logs and reclaimed by the next run or `heph gc`.
- **B — bound the race.** The batch retries only the contended subset, once, after a single short delay for the whole batch. Deterministic; costs one delay on the exit path, and only when something was actually contended.

B is not implemented. The three-platform CI on this PR is the evidence either way — the per-run assertion is what would surface a live race.

### DIFF
--- a/crates/core/src/blocking.rs
+++ b/crates/core/src/blocking.rs
@@ -118,14 +118,31 @@ pub fn backstop(waker: Waker) {
 /// alive past the request that made it.
 ///
 /// Waking early is always sound: a spurious wake costs one poll, and a waiter
-/// still genuinely pending re-registers from that poll.
-pub fn flush_backstop() {
+/// still genuinely pending re-registers from that poll. That re-registration is
+/// the invariant this rests on — [`run`] below and the sqlite cache's
+/// write-behind waiter both register on *every* pending poll. A future
+/// registrant that registers once and relies on the tick would be stranded by a
+/// flush; register per pending poll.
+///
+/// Returns how many registrations were taken, so a caller can tell its own flush
+/// from a tick that happened to land first.
+pub fn flush_backstop() -> usize {
     // Taken before waking, and the lock released first, so a waker that
     // re-registers from inside `wake` cannot deadlock against us.
     let due = std::mem::take(&mut *lock_pending());
+    let n = due.len();
     for waker in due {
-        waker.wake();
+        // Callers reach this from `Drop` on a teardown path. A panicking waker
+        // there would unwind out of a destructor — and abort outright if that
+        // drop is itself already unwinding — so one bad waker is contained
+        // rather than allowed to take the process with it. The panic is still
+        // reported by the default hook before this returns, and this crate has
+        // no `tracing` (it is linked into cdylib plugins, where no subscriber is
+        // ever installed), so there is nothing further to say about it here.
+        // `lock_pending` already contemplates this case for the tick thread.
+        drop(catch_unwind(AssertUnwindSafe(|| waker.wake())));
     }
+    n
 }
 
 /// A waker panicking mid-`wake` would poison the list and strand every later
@@ -305,8 +322,12 @@ mod tests {
         let owned = Arc::new(());
         let started = std::time::Instant::now();
         backstop(futures::task::waker(Arc::new(Owning(Arc::clone(&owned)))));
-        flush_backstop();
+        let taken = flush_backstop();
 
+        assert!(
+            taken >= 1,
+            "the flush itself must have taken our registration, not a tick",
+        );
         assert_eq!(
             Arc::strong_count(&owned),
             1,
@@ -315,6 +336,38 @@ mod tests {
         assert!(
             started.elapsed() < WAKE_BACKSTOP,
             "the release must not have come from a backstop tick",
+        );
+    }
+
+    /// A waiter that is genuinely still pending must survive a flush: it is
+    /// woken, re-polls, finds its job unfinished and re-registers. This is the
+    /// invariant that makes `flush_backstop` safe to call from anywhere.
+    #[test]
+    fn flush_backstop_does_not_strand_a_still_pending_waiter() {
+        let out = futures::executor::block_on(async {
+            let job = run(|| {
+                thread::sleep(WAKE_BACKSTOP / 2);
+                "done"
+            });
+            futures::pin_mut!(job);
+            // Flush repeatedly while the job is still running; each one takes the
+            // waiter's registration out from under it.
+            loop {
+                let mut flushed = 0;
+                let polled = futures::poll!(&mut job);
+                if let Poll::Ready(v) = polled {
+                    break v;
+                }
+                while flushed < 3 {
+                    flush_backstop();
+                    flushed += 1;
+                    thread::sleep(Duration::from_millis(20));
+                }
+            }
+        });
+        assert_eq!(
+            out, "done",
+            "a flushed-but-pending waiter must still finish"
         );
     }
 
--- a/crates/core/src/blocking.rs
+++ b/crates/core/src/blocking.rs
@@ -102,6 +102,32 @@ pub fn backstop(waker: Waker) {
     lock_pending().push(waker);
 }
 
+/// Wake every registered backstop waiter now rather than on the next tick.
+///
+/// A registration outlives the future that made it. [`backstop`] is a
+/// fire-and-forget push: a waiter re-registers on every pending poll and never
+/// unregisters, so once its result arrives its last registration simply sits in
+/// [`PENDING`] until a tick sweeps it — and a `Waker` owns its task, which owns
+/// everything that task's state holds. For up to [`WAKE_BACKSTOP`] after a piece
+/// of work is completely finished, its state is still reachable from this list.
+///
+/// Usually that is invisible: it is bounded, and it costs a little memory. It is
+/// *not* invisible to a caller that must observe those values actually released
+/// — the post-run cache trim needs the request's cache read guards gone before
+/// it can take a write lock, and a stale registration here is enough to keep one
+/// alive past the request that made it.
+///
+/// Waking early is always sound: a spurious wake costs one poll, and a waiter
+/// still genuinely pending re-registers from that poll.
+pub fn flush_backstop() {
+    // Taken before waking, and the lock released first, so a waker that
+    // re-registers from inside `wake` cannot deadlock against us.
+    let due = std::mem::take(&mut *lock_pending());
+    for waker in due {
+        waker.wake();
+    }
+}
+
 /// A waker panicking mid-`wake` would poison the list and strand every later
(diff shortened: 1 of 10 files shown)


raphaelvigee:
Checked one interaction worth recording: an `in_place` target files a **second** revision for the same addr via `store_fixpoint` → `duplicate_cache_revision`, after `execute_and_cache_inner` has already recorded the trim protecting the primary. At `history = 1` that is two revisions and one protected key.

It is already handled, and deliberately: `duplicate_cache_entry` stamps the duplicate with a strictly-newer `created_at_nanos` *specifically* so it outranks the primary in the history trim, and the comment there says in as many words that "the primary is independently protected by the trim, so both survive even at `history = 1`". Sorted newest-first the fixpoint takes the one `keep` slot and the primary is kept by `protect`.

That code was written against a post-write trim that has never actually run. This PR is what makes it load-bearing — and it lands exactly as its author intended.

PR #228, opened 2026-07-28
feat(diag): report worker saturation and keep the in-flight state on disk

raphaelvigee:
Follow-up to #216, driven by a second wedge that #216's inventory narrowed but could not explain.

## The incident

14 fires over 25 minutes, every count byte-identical:

```
open ops     578 result
since last   +0 done, +0 open, over 120s
in-flight    578 result, 455 meta, 123 locked_result, 118 execute_cache, 1 spec
```

Thread dump: 59 threads, all parked, hcore pool on an empty queue. Process entirely idle.

The inventory got us a long way. The chain reads directly — `spec(worker:build_lib)` → `result(worker:_golist)` → `meta(worker:_golist)` — and the frontier is the **123 `locked_result` cells, 99 of them `@heph/fs:file`** source reads, parked past `meta` and short of `Driver::run`. But the paragraph could not say why, because of three gaps.

## 1. `workers` and `codec` were declared and never fed

```
Limiter::observe call sites:
  remote_cache.rs:545   → remote-cache-metadata
  remote_cache.rs:1308  → remote-cache-transfer
  (nothing else)
```

`diag::global()` declares four limiters. Two of them were fed by nothing, so `saturated_for()` was permanently `None`.

That matters most for `workers`: the permit is taken **after** dep resolution and **before** `ExecuteStart` (`execute.rs:53`). A target queued there has no open `execute` span *and* no limits line — invisible in both places it could have shown up, which is exactly where those 99 fs:file cells sit.

Observed before the acquire, so "saturated for 47s" times the wait rather than the moment a permit came free.

## 2. The lost-wake headline keyed on the wrong bit

It required `!has_driver && waiters > 0`. Every cell in this wedge had `driver=true` — an awaiter that is woken and never re-polled keeps occupying the driver slot forever. So the check said nothing across 25 minutes of a completely idle process.

Now keyed on three independent observations: flat delta, cells still incomplete, and no operation of any kind in flight. `Op::Result` is excluded from that last test deliberately — a result span is bookkeeping for "being resolved", not work, and every ancestor holds one open while awaiting children. The stranded sub-case is still called out when present.

```
  Nothing is executing, transferring, or waiting on a lock, 3 cell(s) are
  still incomplete, and nothing moved since the last report. The process is
  idle with work outstanding — a lost wake-up or a dependency cycle, not
  slow work.
  The full list, the wait-for graph and each invocation's next await are in
  `inflight-36381.log` beside this file, refreshed on every report. Re-run with
  `HEPH_DEBUG_MEMOIZER_CYCLE=1 HEPH_PHASE_TRACE=1` to populate the last two,
  and to fail a real cycle instead of hanging on it.
```

## 3. The in-flight state needed someone watching

It was only reachable by sending `SIGQUIT` to a live process. The first incident ended with the process killed before anyone could, and every byte went with it.

The watchdog now writes the same report to `<home>/diag/inflight-<pid>.log` on every fire — uncapped, **truncated** each time so it holds the current state, while the stall log keeps appending the escalation history (which is the part that says "wedged" rather than "slow"). `SIGQUIT` still produces it on demand; it just no longer requires someone to be watching.

Both paths share one renderer (`hmemoizer::render_full_report`), asserted byte-identical by a test — two formats for one thing means working out which is truncated, during an incident.

## Also

`heph_PHASE_TRACE` → `HEPH_PHASE_TRACE`, matching `HEPH_DEBUG_MEMOIZER_CYCLE` and `HEPH_MEMOIZER_STALL_SECS`. Lowercase spelling removed, not aliased.

## Tests

- `a_wedge_is_reported_even_when_every_cell_still_has_a_driver` — the exact shape the old check missed; asserts zero cells are "stranded" and it fires anyway.
- `work_in_flight_retracts_the_idle_claim` / `a_blocked_lock_retracts_the_idle_claim` — the claim is "nothing is happening", so a real op or a blocked lock must retract it.
- `a_saturated_worker_pool_reaches_the_paragraph` — limiter → report → text.
- `the_inflight_log_is_replaced_not_appended`, `the_inflight_report_carries_every_section`, `the_dump_and_the_watchdog_companion_file_render_identically`.
- `phase_trace_is_opt_in_under_the_canonical_name` — asserts the lowercase form no longer works.

## Verification status

Honest accounting, because this host is currently saturated (100+ cargo processes from other worktrees sharing `CARGO_TARGET_DIR`):

- **Green as of the second-to-last commit state**: 34 core + 32 engine unit tests, 2 bin tests, `lint` clean, and a full workspace run at **1229 passed / 0 failed** (excluding `plugingo-e2e`, which fails locally on `No space left on device` staging the Go std toolchain — unrelated, CI has the disk).
- **Not re-run locally**: the final edit, which removes the legacy env-var constant, collapses `phase_trace_from` to a single lookup, and renames one test. I could not get a build slot; CI gates it.

Two `observe()` call sites are also not covered by an automated test — asserting the real acquire site observes correctly needs genuine contention and would be flaky. Both are three-line mirrors of the proven `remote-cache-metadata` pattern.

## Still open

The wedge itself is undiagnosed. Next step is a run with `HEPH_DEBUG_MEMOIZER_CYCLE=1 HEPH_PHASE_TRACE=1` — `execute.rs` is already instrumented with `set_phase("execute:inputs_result_exec")`, `"execute:semaphore_acquire"`, `"execute:sandbox_remove"` (16 sites across the engine), so the phase map will name the exact await each of the 123 is parked at, and cycle detection will fail a real cycle instead of hanging on it.

### DIFF
--- a/crates/core/src/hmemoizer/mod.rs
+++ b/crates/core/src/hmemoizer/mod.rs
@@ -176,12 +176,23 @@ fn format_frame_stack(top: &Arc<Frame>) -> Vec<String> {
 // where each stuck invocation is parked when the hang is on a non-memoizer
 // await (semaphore acquire, fs op, subprocess wait, cache_locally, …).
 //
-// Opt in via `heph_PHASE_TRACE=1`. Disabled by default — `set_phase` and
+// Opt in via `HEPH_PHASE_TRACE=1`. Disabled by default — `set_phase` and
 // `clear_phase` are O(1) early-returns when the flag is off.
+
+/// Spelled like every other knob (`HEPH_DEBUG_MEMOIZER_CYCLE`,
+/// `HEPH_MEMOIZER_STALL_SECS`).
+const PHASE_TRACE_VAR: &str = "HEPH_PHASE_TRACE";
+
+/// The decision, separated from the process environment so it is testable — the
+/// real one caches in a `OnceLock`, so a test that set the var would be at the
+/// mercy of whichever test ran first.
+fn phase_trace_from(get: impl Fn(&str) -> Option<String>) -> bool {
+    get(PHASE_TRACE_VAR).as_deref() == Some("1")
+}
 
 fn phase_trace_enabled() -> bool {
     static FLAG: OnceLock<bool> = OnceLock::new();
-    *FLAG.get_or_init(|| matches!(std::env::var("heph_PHASE_TRACE").as_deref(), Ok("1")))
+    *FLAG.get_or_init(|| phase_trace_from(|name| std::env::var(name).ok()))
 }
 
 static PHASES: OnceLock<Mutex<HashMap<u64, &'static str>>> = OnceLock::new();
@@ -191,7 +202,7 @@ fn phases() -> &'static Mutex<HashMap<u64, &'static str>> {
 }
 
 /// Tag the calling invocation with its next-await `phase`. No-op when
-/// `heph_PHASE_TRACE` is unset or when invoked outside any `once()` scope.
+/// `HEPH_PHASE_TRACE` is unset or when invoked outside any `once()` scope.
 pub fn set_phase(phase: &'static str) {
     if !phase_trace_enabled() {
         return;
@@ -219,7 +230,7 @@ pub fn clear_phase() {
 
 pub fn dump_phases() -> String {
     if !phase_trace_enabled() {
-        return "  (phase trace disabled — set heph_PHASE_TRACE=1)".to_string();
+        return format!("  (phase trace disabled — set {PHASE_TRACE_VAR}=1)");
     }
     let map = phases().lock().expect("phases mutex poisoned");
     if map.is_empty() {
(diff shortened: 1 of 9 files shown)


raphaelvigee:
Added `7f460aed` — **every diagnostic path is now absolute.**

The `SIGQUIT` dump landed at `.heph3/diag/dump-<pid>.txt`, resolved against whatever cwd the process was launched from. That is why the last incident write-up had to say "under the directory you started the build in" rather than a path — a round trip in exactly the situation where the process may already be gone, and useless to anyone handed the file afterwards.

- The dump is pointed at `<home>/diag` by `bootstrap::new_engine` once the home is resolved, so it lands beside the stall log and the in-flight report. Before that it falls back to the launch dir *made absolute* — a hang during startup still has to produce a dump somewhere findable.
- `StallLog` / `InflightLog` absolutize their own paths, so a relative `--home` cannot produce a reported path that resolves differently depending on where the reader stands.
- The stall paragraph prints the in-flight report's full path instead of a bare `inflight-<pid>.log`.

`std::path::absolute`, not `canonicalize` — the diag dir does not exist until the first write, and `canonicalize` fails on a path that is not already there. Symlink resolution is not wanted here anyway; the goal is a path you can paste.

`diag` moves from the binary into the library, because it now has to be reachable from `commands::bootstrap`, which is library code. `main` still installs the signal handler.

New tests: `diagnostic_paths_are_absolute_even_from_a_relative_home` (engine), `the_dump_path_is_absolute` (bin).

### Verification

Resolving the caveat from the PR description: **the env-var rename is now verified** — `core` hmemoizer tests, 34 passed / 0 failed.

For this commit: `cargo check -p heph --lib --bins` clean, and `engine` diag tests 33 passed / 0 failed (includes the new absolute-path assertion). The bin-side `the_dump_path_is_absolute` and a final `lint` pass did not get a build slot — this host is sitting at ~100 concurrent cargo processes from other worktrees sharing `CARGO_TARGET_DIR`. CI covers both.

raphaelvigee:
Verification caveats from the previous comment are now cleared — a build slot freed up and everything ran locally:

| Suite | Result |
|---|---|
| `core` — `hmemoizer::` | 34 passed, 0 failed |
| `engine` — `diag::` | 33 passed, 0 failed |
| `heph --lib` — `diag::` | 3 passed, 0 failed |
| `lint` | clean |

That covers both previously-unrun items: `the_dump_path_is_absolute` and the final clippy/fmt pass.

Note the `diag` tests now run under `-p heph --lib` rather than `--bin heph`, since the module moved into the library in `7f460aed`.

The only thing still not locally exercised is the two `Limiter::observe` call sites, for the reason given in the PR description — asserting them needs real contention and would be flaky. Everything from the limiter through to the rendered line is covered by `a_saturated_worker_pool_reaches_the_paragraph`.

PR #231, opened 2026-07-28
fix(diag): sample limiters when the report is rendered, not at the last acquire

raphaelvigee:
`Limiter::observe` was only ever called from a limiter's own acquire path, so its reading froze the moment nothing acquired any more — which is precisely when the stall watchdog fires.

A wedged build reported:

```
open ops     542 result
limits       workers saturated for 90s
```

with **zero** `execute` spans open. Both cannot be live: the permit is taken immediately before `ExecuteStart` with no await between, so a held permit implies an open execute span. The "90s" was a stamp left by the last acquire attempt before work stopped — indistinguishable from a pool that is still fully held.

Telling those two apart is the whole question at a stall. Permits held means deadlocked on the pool; permits free means wedged elsewhere and the reader should be looking somewhere else entirely.

## Change

`Limiter` gains an optional live gauge, read by `DiagState::saturated` when a report is built and never on the acquire path:

- `workers` → wired to the result semaphore at construction (`engine.rs`)
- `codec` → wired to `CODEC_SLOTS` (`remote_cache.rs`)

Attachment is idempotent, so a hot path can arm it without a separate one-time hook. A limiter with no gauge keeps exactly its old acquire-only behaviour, so nothing starts reporting phantom saturation.

## Tests

- `a_live_gauge_reports_the_present_not_the_last_acquire` — retracts a stale claim when permits come back with no acquire to notice, and reports fresh saturation that happened between reports.
- `a_limiter_without_a_gauge_keeps_its_acquire_only_reading`
- `attaching_a_gauge_twice_keeps_the_first`

## ⚠️ CI will be red for a pre-existing reason

`cargo test -p engine` does **not compile on master** right now, independently of this PR:

```
error[E0308]: mismatched types
   --> crates/engine/src/engine/result.rs:4583:18
    |
4583|         drain_bg(&seed_rs).await;
    |         -------- ^^^^ expected `Arc<RequestState>`, found `&Arc<RequestState>`
note: function defined here
   --> crates/engine/src/engine/result.rs:4148:14
    |
4148|     async fn drain_bg(rs: Arc<crate::engine::request_state::RequestState>) {
```

7 call sites, all added by #224 (`ca4a7000`), against a by-value signature that already existed at that commit. This PR touches only `diag.rs`, `engine.rs` and `remote_cache.rs` — none of the errors are in files it modifies. Worth fixing on its own (it is a 7-character change), but I have left it alone here rather than fold an unrelated repair into this diff.

Verified locally by applying the fix out-of-band: engine `diag::` tests pass with the new gauge tests included.

### DIFF
--- a/crates/engine/src/engine/diag.rs
+++ b/crates/engine/src/engine/diag.rs
@@ -189,6 +189,16 @@ pub struct Limiter {
     pub name: &'static str,
     /// Millis at which this limiter last went from "has permits" to "none", or 0.
     saturated_since_ms: AtomicU64,
+    /// Reads the limiter's *current* free permits, when one has been attached.
+    ///
+    /// Without it a limiter is only ever sampled from its own acquire path, so
+    /// the reading freezes the moment nothing acquires any more — which is
+    /// exactly when the stall watchdog fires. A wedged build then reported
+    /// "workers saturated for 90s" from a stamp left by the last acquire
+    /// attempt, indistinguishable from a pool that is still fully held.
+    /// Telling those two apart is the whole question at a stall: permits held
+    /// means deadlocked on the pool, permits free means wedged elsewhere.
+    gauge: std::sync::OnceLock<Box<dyn Fn() -> usize + Send + Sync>>,
 }
 
 impl Limiter {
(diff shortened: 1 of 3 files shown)


raphaelvigee:
Correcting the verification line in the description — when I wrote it the gauge tests had **not** actually run (an earlier run reported 33 passed, which is the pre-existing count; my three tests were silently absent because the test binary never built).

Now genuinely verified. Applying the `drain_bg` repair locally, out-of-band, to get a compiling test binary:

```
test engine::diag::tests::a_live_gauge_reports_the_present_not_the_last_acquire ... ok
test engine::diag::tests::a_limiter_without_a_gauge_keeps_its_acquire_only_reading ... ok
test engine::diag::tests::attaching_a_gauge_twice_keeps_the_first ... ok
test engine::diag::tests::limiter_reports_how_long_it_has_been_saturated ... ok
test result: ok. 36 passed; 0 failed
```

36 = the 33 that exist on master plus these three. The local `drain_bg` patch was reverted; this branch contains only the three `diag.rs` / `engine.rs` / `remote_cache.rs` files.

`lint` has not been run against this branch — clippy needs the same test build that master currently breaks.

raphaelvigee:
Folded into #237, which needs the same live-gauge mechanism for its worker-permit accounting. Landing one mechanism rather than two.

PR #236, opened 2026-07-28
fix(hmemoizer): never discard a wake destined for the cell's driver

raphaelvigee:
The build still wedges. PID 1873777, this time with the phase trace on:

```
78 @ execute:semaphore_acquire     (and nobody past it)
open ops     542 result            (zero execute spans)
```

Every tokio worker parked, the blocking pool idle on an empty queue, 1163 memoizer cells incomplete. **Nobody holds a worker permit through normal execution** — there are no open `execute` spans at all — yet 78 targets sit on `execute`'s semaphore and never advance.

tokio's `Semaphore` *assigns* a permit to the first queued waiter on release and wakes it. Drop that wake and the permit is stranded inside an `Acquire` future nobody polls again: permits gone, nothing running, everyone queued. That is exactly the state above.

## Where the wake is dropped

`Cell::wake_by_ref` took the driver's waker **out** of the slab, and on the next wake found the slot empty. A `driver_awake` latch then said the driver "already owes a re-poll", and the wake was discarded:

```rust
let targeted = wakers.take(driver);          // ← removes the registration
if let Some(waker) = targeted { ...; return }

if driver != NO_DRIVER && cell.driver_awake.load(Acquire) {
    return;                                   // ← swallowed
}
```

The justifying comment: *"polling the inner future re-reads its state, so nothing is lost by staying quiet."*

True for a future whose readiness can be **re-read** — a `oneshot` still holding its value. False for one that takes a wake as a **one-shot handoff**, which is what a semaphore acquire becomes once tokio has assigned it a permit.

## Change

The driver is woken through a *clone* of its registration (`Wakers::peek`) instead of taking it. The slab keeps its waker until it re-registers or goes away, so a second wake always has somewhere to land — and the swallow branch, plus `driver_awake` and the now-unused `Wakers::take`, are gone.

**The property the module exists for is untouched**: one task is woken per inner wake, not W. Waking an already-scheduled task is a cheap no-op in tokio. `inner_wake_reaches_only_the_driver` still asserts the parked awaiters get zero wakes across 100 inner wakes — only the driver's expected count changes from 1 to 100.

This is the same defect #232 chased on `hcore::blocking::run`. That fix armed a backstop for one await; the bug moved to the next await with the same shape. This removes the class rather than the instance.

## Tests

`a_second_wake_before_the_driver_re_polls_is_still_delivered` is the regression test. Restoring `take()` fails it:

```
assertion `left == right` failed: the second wake was swallowed — a one-shot handoff is lost forever
  left: 1
 right: 2
```

Two existing tests encoded the old contract and are updated:
- `inner_wake_reaches_only_the_driver` — driver now woken on every wake; the anti-storm assertion (parked awaiters at 0) is unchanged.
- `inner_wake_falls_back_to_everyone_when_the_driver_slot_is_empty` — with clone-not-take, "registered driver, empty slot" is no longer reachable, so it now simulates a genuinely absent driver (`driver = NO_DRIVER`) and asserts the fan-out still happens.

Local: core lib 89 passed, engine lib 395 passed, full `e2e` green, `lint` clean. Rebased on `e349152c`.

## Worth a second opinion

This is a hot path — every `once()` awaiter goes through it — and the change trades a `Waker` clone per inner wake against the removed `take`. I expect it in the noise, but I have not measured it, and `perf-measurement` should confirm rather than my asserting it. The wake *count* is unchanged; it is the clone that is new.

### DIFF
--- a/crates/core/src/hmemoizer/cell.rs
+++ b/crates/core/src/hmemoizer/cell.rs
@@ -21,6 +21,27 @@
 //! the inner future, identified by its slab key in [`Cell::driver`]. Everyone else
 //! is woken exactly once, when the value is ready.
 //!
+//! ## A wake is never discarded
+//!
+//! The driver is woken through a *clone* of its registration, never by taking it.
+//! An earlier version took the waker out of the slab and, on the next wake,
+//! found the slot empty; a `driver_awake` latch then told it the driver "already
+//! owes a re-poll", and the wake was dropped on the reasoning that re-polling the
+//! inner future re-reads its state anyway.
+//!
+//! That reasoning holds for a future whose readiness can be re-read — a `oneshot`
+//! still holding its value — and is false for one that takes a wake as a
+//! **one-shot handoff**. `tokio::sync::Semaphore` is the latter: on release it
+//! *assigns* a permit to the first queued waiter and wakes it. Swallow that wake
+//! and the permit is stranded inside an `Acquire` future nobody will poll again.
+//! A wedged build showed exactly that — every worker permit gone, 78 targets
+//! queued on `execute`'s semaphore, and not one thread doing anything.
+//!
+//! So the slab keeps the driver's waker until it re-registers or goes away, and a
+//! wake with a live driver is always delivered. The property that matters is
+//! unchanged: one task is woken, not W. Waking a task that is already scheduled
+//! is a cheap no-op.
+//!
 //! ## What is kept, and why
 //!
 //! - **Election is per-poll, never "first awaiter wins".** Drivership is held only
@@ -56,7 +77,7 @@ use futures::future::BoxFuture;
 use futures::task::{ArcWake, waker_ref};
 use std::future::Future;
 use std::pin::Pin;
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex, MutexGuard, Once, OnceLock};
 use std::task::{Context, Poll, Waker};
 use std::thread;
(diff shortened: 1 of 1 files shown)


raphaelvigee:
Moved into stack #239 as the top layer (base is now #238's branch), because a new wedge shows this PR is not optional — it is the fix for that instance.

PID 2055961, built from the stack **without** this change:

```
open ops     4665 result, 12 execute (oldest 163s)
workers      12 max, 0 free, 12 running     ← unaccounted = 0
12 @ execute:sandbox_remove
```

The worker-permit accounting added in #237 settles what three earlier dumps could not: `unaccounted = 0`, so no permit is stranded. The pool is genuinely busy with 12 targets that really are running — and all 12 are parked at `execute:sandbox_remove`, which is one await, `hcore::blocking::run`.

Meanwhile the thread dump shows all 24 blocking-pool threads idle on an empty queue (`blocking.rs:266`, `for job in rx.iter()`) and **zero** threads in `remove_dir_all`. The jobs finished. Twelve tasks are waiting for results that already arrived, holding every permit.

The chain, with each link now evidenced rather than inferred:

1. The job completes, sends its oneshot, and wakes the **cell's** waker.
2. `Cell::wake_by_ref` takes the driver's waker out of the slab and sets `driver_awake`.
3. `hcore::blocking`'s backstop retains its registration (#232, #234) and re-wakes the cell every 250ms.
4. **Every one of those ticks is swallowed** — driver set, `driver_awake` true, slot empty.
5. Never re-polled. Permit held for the life of the process.

So the backstop added in #232 is structurally defeated by this swallow: it fires correctly, forever, into a cell that discards it. That is why retention alone did not fix anything.

It also explains why the build that *did* include this PR (PID 1910570) wedged in a different place — 0 at `sandbox_remove`, 107 at the semaphore. This fixes that strand; #238 fixes the other. Both are needed, which is why they are now stacked rather than competing.

Rebased onto #238's branch; core lib 89 passed, engine lib 406 passed, `lint` clean.

PR #237, opened 2026-07-28
feat(diag): account for worker permits, and name abandoned cells

raphaelvigee:
Diagnosis only — nothing about how the engine runs changes. Two prior fixes (#232, #236) were aimed at inferred causes and neither held, so this measures before touching anything else.

## The unresolvable reading

Three wedges have now ended at the same place. From the latest (PID 1910570, built from #236 so that fix *was* in):

```
open ops     585 result            ← zero execute spans
limits       workers saturated for 340s
107 @ execute:semaphore_acquire    ← and nobody past it
```

Every worker permit gone, 107 targets queued on the semaphore, and **nothing executing** — no open execute span, no `waitpid`/`pluginexec`/subprocess frame on any of the 59 threads.

"The pool is busy" and "the pool has been leaked away" produce identical output today. They want opposite investigations.

## The arithmetic that separates them

A tokio `Semaphore` hands a released permit to the first queued waiter and wakes it. So a permit granted to a future that is never polled again is **spent and held by nobody**: the pool reports it taken, and no target reports running.

`capacity - free - running` names exactly that population.

- The engine registers its pool with the diag table: permit count, plus a live reader of the free count.
- `execute` marks a permit *running* only once `acquire()` has **returned** — a permit handed to a waiter that is never polled never reaches that line, which is precisely what makes the two counts disagree.

```
  workers      12 max, 0 free, 3 running, 9 unaccounted
```

The last term prints only when non-zero, so a healthy busy pool stays quiet about it.

## Abandoned cells

The inventory now separates **abandoned** cells (no awaiters at all, never completed) from stranded ones (waiters, no driver).

A cell keeps its in-flight future when an awaiter is dropped — deliberately, so another awaiter can take over. But when the *last* awaiter goes for good (fail-fast drops every sibling on the first error; Ctrl-C drops them wholesale) there is no successor, and that future is parked for good **while keeping its place in whatever queue it was waiting on**.

The last dump had 42 of them, visible only by manually reading a `waiters=0 driver=false` column. They are now counted, marked `ABANDONED` in the listing, and sorted near the top.

## What this will tell us

On the next wedge, one of two things:

- **`unaccounted > 0`** → permits are stranded in futures nobody polls, and the abandoned-cell count says where they went. That confirms the hypothesis and points the fix at cancelling abandoned computations.
- **`unaccounted == 0`** → the pool is genuinely occupied and my reading of the last three dumps is wrong; the investigation moves to what those running targets are blocked on.

Either way it replaces an inference chain with a measurement.

## Tests

- `unaccounted_permits_are_named` — the measurement itself, end to end through the rendered paragraph.
- `a_fully_busy_pool_reports_no_unaccounted_permits` — a healthy full pool must not read as leaking.
- `a_released_permit_stops_being_counted` — the guard is tied to the permit's scope, so the count cannot drift; a drifting count would make the arithmetic worthless.
- `an_unregistered_pool_reports_nothing` — no invented line before the engine registers.

Local: core lib 88 passed, engine lib 399 passed, full `e2e` green (10 suites), `lint` clean. Rebased on current master.

## Note on #231

#231 adds a live gauge to `Limiter` for the same underlying reason (an acquire-time stamp is stale exactly when a stall report is rendered). This PR is independent and self-contained — it adds its own live reader for the worker pool specifically. If you would rather land one mechanism, I am happy to fold #231 into this and close it.

### DIFF
--- a/crates/core/src/hmemoizer/mod.rs
+++ b/crates/core/src/hmemoizer/mod.rs
@@ -461,6 +461,23 @@ impl StuckCell {
     pub fn is_stranded(&self) -> bool {
         !self.has_driver && self.waiters.is_some_and(|n| n > 0)
     }
+
+    /// Nobody is awaiting this cell at all, and it never finished.
+    ///
+    /// The cell still *holds* its in-flight future — that is deliberate, so an
+    /// awaiter dropped between polls can be replaced by another. But when the
+    /// last awaiter goes for good (fail-fast drops every sibling on the first
+    /// error; Ctrl-C drops them wholesale) there is no replacement coming, and
+    /// nothing will ever poll that future again.
+    ///
+    /// It is not inert while it sits there. A parked future keeps whatever it
+    /// was holding, and keeps its place in whatever queue it was waiting on — so
+    /// an abandoned computation can still be handed a worker permit it will
+    /// never use and never give back. Counting these is how that becomes
+    /// visible instead of inferred.
+    pub fn is_abandoned(&self) -> bool {
+        !self.has_driver && self.waiters == Some(0)
+    }
 }
 
 /// Type-erased handle on one live `Memoizer`'s cache.
@@ -542,6 +559,7 @@ pub fn inventory() -> Vec<StuckCell> {
     out.sort_by(|a, b| {
         b.is_stranded()
             .cmp(&a.is_stranded())
+            .then(b.is_abandoned().cmp(&a.is_abandoned()))
             .then(b.waiters.unwrap_or(0).cmp(&a.waiters.unwrap_or(0)))
             .then(a.tag.cmp(b.tag))
             .then(a.key.cmp(&b.key))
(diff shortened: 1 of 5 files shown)


raphaelvigee:
Folded #231 into this branch — the  live gauge is now here too, and #231 is closed. Both readers hit the same semaphore: the permit accounting (`N max, N free, N running`) and the `workers` limiter's saturation age, each sampled when the report is rendered rather than at the last acquire.

Also worth recording here: I ran the reachability analysis on the existing dump (PID 1910570), walking the wait-for graph upward from every `execute_cache` cell queued on the semaphore:

```
execute_cache cells queued on the semaphore: 99
  live-rooted (a real caller still wants the result): 78
  dead-rooted (every root is an abandoned cell):      21
```

The dump has 24 blocking-pool threads and `pool_size() = cores * 2`, so 12 cores and `max_workers = 12`. **21 dead-rooted waiters ≥ 12 permits** — every permit can be absorbed by a chain nobody will ever poll, which is exactly the observed state (0 free, nothing running, 78 live waiters queued forever).

So this PR is now confirmatory rather than exploratory: the diagnosis already holds without it. Still worth landing — it turns a 40-line offline script into a line in the paragraph, and the abandoned-cell count is worth having permanently.

PR #241, opened 2026-07-28
fix(hmemoizer): cancel a computation once nobody is awaiting it

raphaelvigee:
Stacked on #236. **This is the fix for the wedge** — the three PRs below it made it diagnosable; none of them fixed it.

## The failure

Twice reproduced (PIDs 2055961, 2082719), structurally identical:

```
open ops     3261 result, 12 execute (oldest 269s)
workers      12 max, 0 free, 12 running     ← unaccounted = 0
12 @ execute:sandbox_remove   774 @ execute:semaphore_acquire
abandoned    516 cell(s) have no awaiters left
```

Every permit held, ~800 targets queued, and **nothing executing**: the blocking pool parked on an *empty queue*, zero threads in filesystem work or a subprocess wait, all 59 threads in futex/epoll/nanosleep. The jobs had finished; their results were never read.

## Mechanism

A cell keeps its in-flight future while awaiters come and go — deliberately, so one dropped between polls can be replaced. When the **last** awaiter goes there is no successor, and fail-fast produces that in bulk (1,135 and 1,949 failures in the two runs). The future stays parked holding a worker permit, and its wake routes to a driver living inside the parked future of an abandoned ancestor, where `wake_by_ref` wakes an empty waker set. The completion wake and ~1,000 backstop ticks over 269s all evaporate.

Proven verbatim from the dump for e.g. `…/idps/errors.go`:

```
[result] …errors.go…  waiters=0 driver=false ABANDONED   (4/4 rows)
 → owner 1380857 → [locked_result] → owner 1380865
 → [execute_cache] → inv 1380866 @ execute:sandbox_remove   ← holds a permit
```

`StuckCell::is_abandoned`'s doc comment (added in #237) predicted this verbatim.

## Fix

Cancel a computation once nobody wants it. `Memoizer::process` counts interest under the cache lock; when the last holder goes, the key is evicted and the future dropped. The drop cascades — inner awaits abdicate, the execute future goes, `_permit` releases, queued acquires leave the semaphore, backstop registrations disarm.

**Interest is counted explicitly and cannot be `Arc::strong_count`.** The cell *is* its own waker (`waker_ref(cell)`), so every parked leaf holds a strong clone — a `oneshot`, a queued semaphore acquire, a backstop registration, every child cell's waker slab. A `strong_count` gate was tried first and was **dead on arrival in production**; only `futures::future::pending()`, which stores no waker, made it look correct. Two tests pin that lesson.

**The locked re-check is `interest != 0 || is_done()`, and it is exact, not advisory:** under the cache lock, completion requires a poll → a poll requires a live `Await` → a live `Await` requires interest → interest is only raised under that lock. Without the `is_done()` half, a joiner that arrived *and completed* inside the window would have its memoized value evicted — a double build.

Also breaks a reference cycle that leaked whole request states (future → `Arc<RequestState>` → memoizer → cell → future). The wedged inventory showed **3,635 duplicate keys of 3,674** — several dead requests still resident.

## Sandbox lifecycle, forced by making cancellation routine

- **`sandbox_cleaner::generation`** gates each path. The claim (increment/compare, no I/O) is taken inline before the removal is queued; the walk runs under a second lock held only by dedicated OS threads and declines if the path was reclaimed. Two locks because the claim runs on a tokio worker and must never block behind a multi-second `remove_dir_all`. Without the decline, a cancelled execute's queued removal could delete a **successor's live sandbox mid-run** — silent output corruption. (Caught by review; red test now green.)
- **`SandboxTeardown`** resolves exactly once per claimed execute: `complete()` enqueues the bridge cleanup; a bare drop — cancellation, any `?`, unwind — enqueues a generation-checked reclaim, counted in `bg_pending` so shutdown drains it. The old `defer!` fired at drop time *unguarded*, and `gc` has **no sandbox sweep**, so without this a cancelled target never re-executed leaked its tree forever.

`stacker::maybe_grow` wraps the cascade drop — recursive drop glue, one frame-set per level; the deep-cascade test dies with a stack overflow without it (verified by two agents independently). Already a workspace dep, pure Rust + psm on all three targets, cancellation path only.

## Verification

The production repro — three real memoizer layers, a **real `Semaphore` permit** at the leaf, parked on a `oneshot` (stores its waker; `pending()` is explicitly forbidden and asserted against) — fails pre-fix:

```
assertion `left == right` failed: the worker permit must return to the semaphore
  left: 0   right: 1
```

core **103**, engine **416**, e2e **10 suites / 67 tests**, lint clean. All reproduced independently.

## Process

Diagnosed and reviewed by two independent Fable 5 agents over three adversarial rounds, plus a 9-agent read-only pass that re-derived the mechanism from the raw dumps. **Both of my earlier designs were killed by review** — the `strong_count` gate and the unconditional `remove_stale` — and the grounding pass corrected my own claim that all 12 holders sat under abandoned cells (only 2/12 and 3/12 are provable; my reachability analysis was unsound because the wait-graph is deduped by `(tag, key_hash)` while the inventory is per-instance).

## Known residuals, recorded not hidden

- **skip-lock (`@heph/fs`) concurrency**: two live same-addr executes can overlap and their walks can still race. Pre-existing, and *strictly less destructive* now (previously both walks deleted unconditionally).
- **Cross-process stragglers**: the generation registry is process-local. Pre-existing, but the volume of destructive detached jobs is now higher. Recommended follow-up: the reclaim try-flocks the addr and declines if held.
- Reviewer recommends splitting this into two stacked PRs (sandbox hardening first, memoizer cancellation on top) so the behaviourally riskier half has its own revert line. I kept it as one per the request; happy to split on request — the reverse order is *not* available, since cancellation without the generation work makes the corruption race live.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -847,6 +847,7 @@ dependencies = [
  "libc",
  "rustc-hash",
  "serde",
+ "stacker",
  "tar",
  "tempfile",
  "tokio",
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -16,6 +16,7 @@ enclose = "1"
 xxhash-rust = { version = "0.8.15", features = ["xxh3", "std"] }
 rustc-hash = "2"
 libc = "0.2"
+stacker = "0.1"
 tokio = { version = "1.52.3", features = ["rt", "rt-multi-thread", "macros", "sync", "time", "test-util"] }
 
 [dev-dependencies]
--- a/crates/core/src/blocking.rs
+++ b/crates/core/src/blocking.rs
@@ -608,6 +608,78 @@ mod tests {
         assert_eq!(futures::executor::block_on(fut), 11);
     }
 
+    /// The production wedge's leaf, driven through the real machinery: a
+    /// memoized chain whose innermost computation holds a semaphore permit and
+    /// parks inside [`run`], with a live backstop registration — then the
+    /// chain's only awaiter is dropped, as fail-fast does.
+    ///
+    /// Everything the parked leaf holds must come back: the permit returns to
+    /// the semaphore, and the backstop registration is disarmed (a `tick()`
+    /// wakes nobody). Pre-fix, the memoizer cell retained the whole chain, the
+    /// permit stayed captured, and the backstop re-woke the abandoned cell
+    /// every 250ms forever — the dumps show ~1000 such ticks over 269s.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn an_abandoned_memoized_wait_releases_its_permit_and_its_backstop() {
+        use crate::hmemoizer::Memoizer;
+
+        let _exclusive = exclusive();
+
+        let mem_result: Arc<Memoizer<String, u32>> = Arc::new(Memoizer::with_tag("bk-result"));
+        let mem_execute: Arc<Memoizer<String, u32>> =
+            Arc::new(Memoizer::with_tag("bk-execute_cache"));
+        let permits = Arc::new(tokio::sync::Semaphore::new(1));
+
+        // The job blocks until released, so the wait is genuinely pending (and
+        // the backstop genuinely armed) when the abandonment happens.
+        let (release_tx, release_rx) = crossbeam_channel::bounded::<()>(1);
+
+        let mut outer = Box::pin(mem_result.process("//pkg:tgt".to_string(), {
+            let mem_execute = Arc::clone(&mem_execute);
+            let permits = Arc::clone(&permits);
+            move || async move {
+                mem_execute
+                    .process("//pkg:tgt".to_string(), move || async move {
+                        let _permit = permits
+                            .acquire_owned()
+                            .await
+                            .expect("semaphore is never closed");
+                        run(move || {
+                            drop(release_rx.recv());
+                            7
+                        })
+                        .await
+                    })
+                    .await
+            }
+        }));
+        assert!(
+            futures::poll!(&mut outer).is_pending(),
+            "the chain must park inside blocking::run"
+        );
+        assert_eq!(
+            permits.available_permits(),
+            0,
+            "permit captured by the wait"
+        );
+
+        // Fail-fast: the only awaiter goes away while the job is still running.
+        drop(outer);
+
+        assert_eq!(
+            permits.available_permits(),
+            1,
+            "abandoning the chain must release the permit held across the blocking wait"
+        );
+        // The permit coming back proves the innermost future was dropped, and
+        // dropping it drops its `Backstop` — whose registration removal is
+        // proven deterministically by `a_registration_is_released_when_its_wait_ends`.
+        // Asserting `tick() == 0` here instead would race the non-exclusive
+        // tests in this module, which arm registrations of their own.
+
+        // Let the job finish; its answer goes nowhere, and that is fine.
+        release_tx.send(()).expect("job is waiting on this");
+    }
+
     /// Many jobs at once all complete, exercising the queue past the pool size.
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn queues_beyond_the_pool_size() {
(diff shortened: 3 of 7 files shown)


raphaelvigee:
Rebased onto #242 (the sandbox layer), which now carries the `SandboxTeardown` work that was split out of here. This PR is now **just the memoizer fix**: `crates/core/*` plus the `worker_pool` test and the stall-report version line.

Since the original description was written, CI found a regression in the sandbox layer and it has been fixed there — see #242. Nothing in this PR changed as a result; it is the same memoizer diff, rebased.

Current state, verified on the stacked HEAD:

- core **103**, engine **417**, e2e **10 suites / 67 tests**, lint clean
- the production repro still fails pre-fix with `left: 0 / right: 1` on *the worker permit must return to the semaphore*
- the sandbox layer's regression test passes **3/3 in isolation** on this branch

Review status: four adversarial rounds. Rounds 1–3 killed three designs of mine (`Arc::strong_count` as the interest gate, unconditional `remove_stale`, and wake-routing before that); round 4 cleared the final gate with no blocker.

raphaelvigee:
## Correction: the causal story in the original description was wrong

The user corrected the run conditions — **no Ctrl-C, no panic, `--fail-fast` off**, a bare `heph r lint //...`, 15 failing targets (not the 1,135 the stall log reports as "unsuccessful", which counts every dependent that inherited an error). I had attributed abandonment to "fail-fast drops every sibling, Ctrl-C drops them wholesale". Both are real paths; neither happened here. The commit message is amended.

**The actual trigger is the ordinary in-place codegen path, and it needs no failure at all.**

`new_hash_only_state` hardcodes `fail_fast = true` and builds a *fresh* set of memoizers. Its only production callers are `check_in_place_inputs_unchanged` and `maybe_store_fixpoint`, both gated on `is_top` + `CodegenMode::InPlace` — so every top-level in-place codegen target that executes spins up one or two. `:lint` is `go_lint_fix`, declared `CodegenMode::InPlace`, so on this run that is *every* target.

A hash-only request routinely answers `HashUnknownError` **by design** — it refuses to build, and the engine documents that and swallows it with a `debug!`. But it is still an `Err`, and it still reaches the hardcoded-fail-fast `try_join_all` in `inputs_result_meta`, which drops every sibling in flight at that level and every level above. Each dropped sibling's `mem_result` cell is abandoned.

The dumps confirm it exactly:
- **100%** of abandoned cells are `[result]` with key `(addr, OutputMatcher::None, is_top=false)` — and the only production site constructing `OutputMatcher::None` is `inputs_result_meta`
- driverless by tag: `result` 523, `meta` 0, `locked_result` 0, `execute_cache` 0 — the exact signature of dropping a `result_addr` future
- every `is_top` cell is a `//mgmt/go/...:lint`
- 20-25 live `RequestState`s resident, matching the nested-request count

**This raises the severity.** A completely healthy codegen build leaks a whole request state and a worker permit per in-place target — no user error, no interruption, nothing in the log. The right framing is *a per-run resource leak on the ordinary codegen path*, not "cleanup after Ctrl-C".

## One defect fixed as a result

Review found `take_future`'s `WouldBlock` arm — asserted unreachable — is reachable via the two-cancellations interleaving `cancel_abandoned`'s own doc enumerates: both cancellers observe their own zero crossing and both call it with no lock held, so they race on the mutex. The `debug_assert!(false)` fires inside `Drop` glue, which **aborts** if that drop is unwinding. Now a blocking `lock()` — the only contender is another `take_future`, holding the guard for one `Option::take` and acquiring nothing else, so it cannot deadlock.

## Scope correction

This collects cells whose **interest** reaches zero. It does not collect *stranded* cells (waiters, no driver — interest non-zero by construction); those are the lost-wake case owned by #236. A stranded cell also pins its entire descendant subtree, so the cascade stops there — the residual is 7 *subtrees*, not 7 cells, and this PR's value is contingent on #236 having genuinely eliminated stranding.

Also worth noting for reading the next dump: post-fix, an interest-zero in-flight cell is evicted before the inventory can walk it, so the `abandoned` line should read ≈0. A standing nonzero value there is a regression signal.

Re-verified after the `take_future` change: core **103**, engine **417**, e2e **10 suites / 67 tests**, lint clean.

PR #249, opened 2026-07-28
fix(engine): retry a deferred trim that lost its lock, once

raphaelvigee:
P6.3 — resolves the open question left behind by #222.

`hermeticity` flagged that `flush_backstop()` → `try_write` has no happens-before edge: whether the `Arc` holding the addr's riding cache read is dropped before the cleaner thread asks for the write lock depends on who happens to be awaiting. The user chose **option B** — hedge it with a bounded retry — over accepting best-effort.

## What it does

`Engine::run_trim_batch_with_delay` owns the drain: three bounded attempts per target, **at most one sleep for the whole batch**.

1. one pass over the batch;
2. a flush, then an **immediate re-probe** of whatever came back contended;
3. only if that still loses: sleep once (`TRIM_RETRY_DELAY`, 25ms), flush once more, final attempt.

Step 2 was not in the brief and is the most useful part. `flush_backstop` *releases* the `Waker`s it takes (#234), and for the shape that matters — an `Arc<hmemoizer::Cell>` whose memoized `mem_locked_result` value is the riding read — dropping the last one tears the cell down inline, on the flushing thread, before the call returns. When that was the only owner the guard is already gone and a sleep would be pure exit latency. **The delay is charged only to a run whose contention outlives a flush**, not to every run that saw contention.

Never a loop, and still not a blocking `write` — #222's reasoning stands.

## What it does not claim

A hedge, not a synchronization edge. Another live `Arc` (an in-flight upload, a task mid-poll) keeps the read alive and the release then lands unordered with respect to our `try_write`. What is unconditional: two more attempts per contended addr, one bounded sleep per batch, no lock held across it, every deletion decided under a freshly acquired exclusive guard with `written_hashin` still protected.

## For the reviewer

- **`TrimOutcome`** is now `Settled { removed, bytes } | Contended | Failed`. `Contended` is `Ok(None)` and nothing else; #219's within-budget early return is `Settled`. `Err` is never retried — not because it is permanent, but because the retry's only mechanism is waiting for a *holder* to let go and an `Err` is not evidence one exists.
- **Three flushes**, each load-bearing and independently red-checked. Their invariant — no result-lock guard held on this thread across a flush — is written down at the site.
- **Diagnosability**: the batch reports `batch / retried / delayed / still_contended / failed / removed / bytes / flushed`. `flushed` says whether the hedge is doing anything at all. `still_contended == batch` warns: that is the original bug returning, and at `debug` it would return as quietly as it arrived.
- **`bg_pending` is held across the retries** (they run inside the cleaner job), pinned by its own test.

## Verification

Every mechanism verified red on its own: deleting the delayed attempt, each of the three flushes, the immediate re-probe, `protect`, and reclassifying `Err` as `Contended` each fail a different test. 20 back-to-back engine-suite runs clean; `lint` (both feature sets) and `fmt` clean; engine lib (407) and `crates/e2e` green.

One flake was found and fixed during the soak — a test writing cache entries without barriering the sqlite write, so the batch enumerated one revision and settled within budget instead of contending.

## Needs a decision from you

**The 25ms constant is Linux-shaped.** The hedge releases via `waker.wake()`, and a dropped cross-thread wake on darwin is the hazard `hcore::blocking`'s backstop exists for — if a flush's own wake is dropped there, no wait short of the 250ms `WAKE_BACKSTOP` tick recovers it. Semantics stay uniform across all three supported targets (no `cfg`), so this is not silent divergence, but a per-OS constant is an option and per CLAUDE.md that call is yours, not mine. Flagged, not taken.

## Overruled, with reasons

- **Capping the retried subset** (`feature-quality`): later passes are subsets of the first, so the hedge costs ≤2× work the run already pays and the sleep is O(1) per batch. A cap would preferentially disable enforcement on the largest builds.
- **`holder_pid` gating**: the pid is stale in both directions, so a dead foreign pid would make us skip the delay on our own unlanded guard. Logged (only when it is not us), not acted on.

Review board: `hermeticity` HERMETIC, `code-quality` PASS, `feature-quality` blockers addressed (injectable delay; `bg_pending` pinned).

### DIFF
--- a/crates/engine/src/engine/gc.rs
+++ b/crates/engine/src/engine/gc.rs
@@ -13,7 +13,8 @@
 //!   the revision that was just written. Deferred to the end of the request
 //!   that wrote the revision (see `RequestState::defer_trim`) — a request holds
 //!   a read on every addr it resolved, so a trim run inline could never take
-//!   the write lock.
+//!   the write lock. The whole request's batch is drained by
+//!   [`Engine::run_trim_batch_with_delay`], which retries the contended subset.
 //!
 //! Revision recency comes from each group's `manifest-v1.borsh`
 //! (`created_at_nanos`); the full artifact name list to delete comes from the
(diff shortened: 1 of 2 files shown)


raphaelvigee:
Rebased onto `dfc4e5cc` (#241). Clean rebase, no conflicts — `enqueue` has no `Lane` parameter yet, so the mechanical clash with #226 is still ahead of both of us; whichever lands first forces the other to rebase.

**Re-checked against #236/#241, since both are adjacent to this PR's release-edge reasoning.**

They make the release edge *more* reliable, which is the desired direction: #236 stops a wake destined for a cell's driver being discarded, and #241 evicts an abandoned cell and drops its in-flight future, disarming its backstop registration. #241 retires one of the owners this hedge was explicitly written against — a cell retained after a `fail_fast` fanout dropped its last awaiter, pinning the riding read indefinitely. The `TRIM_RETRY_DELAY` doc named that case; it no longer does, and says what changed instead. That is the only code movement in this push (comment-only).

So step 3 (the delayed attempt) should be reached *less* often now. That costs nothing, because the free re-probe in step 2 absorbs the difference — the delay was already charged only to contention that outlives a flush.

**It does not make the tests vacuous**, which was the specific risk. None of them routes a guard through a memoizer: they arm a `Backstop` directly and hold a real `ResultWriteGuard`, so each pins its phase deterministically regardless of what the memoizer does. Confirmed empirically rather than by argument — all seven mutations were re-run against the new base and each still fails the same test as before:

| mutation | test that goes red |
|---|---|
| delete the delayed attempt | `..._retries_the_contended_subset_and_reclaims`, `..._flushes_again_after_the_delay` |
| delete flush #1 (pre-first-pass) | `..._flushes_before_the_first_pass` (+2) |
| delete flush #2 (pre-re-probe) | `..._reprobes_after_a_flush_before_paying_the_delay` (+1) |
| delete flush #3 (post-sleep) | `..._flushes_again_after_the_delay` |
| delete the immediate re-probe | `..._reprobes_after_a_flush_before_paying_the_delay` |
| drop `protect` | `..._retry_sees_revisions_written_during_the_delay` |
| `Err` reclassified as `Contended` | `..._does_not_retry_an_unaskable_lock` |

Also green locally: 429 engine lib tests, 15 back-to-back soak runs of the engine suite, workspace-root clippy on both feature sets, `cargo fmt --check`.

The 25 ms constant being Linux-shaped remains an open decision for you — unchanged and still flagged in the description.

PR #259, opened 2026-07-29
fix(ci): make `lint` cover the whole workspace, and fix the 323 errors it exposes

raphaelvigee:
## The gate was not linting the workspace, and it could green on code that does not compile

`lint` ran `cargo clippy --all-targets --locked -- -D warnings` from the repo root.

The repo root **is itself a package** (`heph`), so cargo's default selection is that package **and its dependency graph**. Two things fall outside it, and both were verified by injecting a deliberate error and re-running rather than reasoned from the flag name:

1. **Members the root binary does not depend on are never compiled at all** — `plugin-go`, `plugin-gha`, both cdylibs, `plugin-abi`/`sdk`/`stabby`, `testkit`, and the `e2e` crates.
2. **No member's test targets are built**, whether the member is in the graph or not.

> **Correction to an earlier version of this description.** It claimed member libs were "compiled with zero lints applied". That is **false**, and the demonstration below is what caught it: injecting `v[0]` into `crates/walk`'s lib *does* fail the old invocation, because cargo passes a package's own `[lints]` to rustc regardless of primary-package status. The hole is narrower than that — "never compiled" and "tests never compiled" — which is precisely why `plugin-go`'s two pre-existing `indexing_slicing` errors were invisible while `walk`'s would not have been.

### Break 2 is Break 1

The two reported breaks are the same bug. On 2026-07-29, run [30441851357](https://github.com/hephbuild/heph/actions/runs/30441851357) on master `dfc5e5de`:

| Job | Result |
|---|---|
| **Lint** | success |
| Test linux/amd64 | FAIL `error[E0046]: not all trait items implemented, missing: exists_committed` |
| Test linux/arm64 | FAIL same |
| Test darwin/arm64 | FAIL same |

The missing impl was inside `mod tests` in `crates/engine/src/engine/result.rs` (fixed later by #252) — a **member unit-test target**, which the gate provably never compiled. `Lint` was not wrong about what it checked; it just never checked that.

Ruled out, not assumed:

- **Not sccache.** That same run's own `sccache --show-stats` step prints `Cache timeouts 0`, and the clippy pass finished in 1m35s.
- **Not a masked exit code.** The script is a plain `&&` chain with no pipes anywhere, run under `devenv shell bash -- -e`.

### A third, quieter break

The `--all-features` pass was a **0.30 s no-op**:

```
2026-07-29T10:04:15.9  > clippy --all-features
2026-07-29T10:04:16.2      Finished `dev` profile ... in 0.30s
```

The root package's only feature (`fuse-sandbox`) is already default, and the feature-gated code the comment promises to cover (`plugin-stabby/host`, `plugin-sdk/stabby`) lives in members that were never selected. It had never once compiled that code.

## The fix

```diff
-cargo clippy --all-targets --locked -- -D warnings
-cargo clippy --all-targets --all-features --locked -- -D warnings
+cargo clippy --workspace --all-targets --locked -- -D warnings
+cargo clippy --workspace --all-targets --all-features --locked -- -D warnings
```

plus `set -euo pipefail`, and a comment stating what each flag buys, what it cost to omit, and the gaps that remain.

**`fix` had the identical defect** and is fixed in the same PR: `cargo fix --allow-dirty` also reached only the root package, had no `--all-targets`, and applied *rustc's* suggestions rather than clippy's — which are what the gate fails on. Now `cargo clippy --fix --workspace --all-targets`. Shipping `lint` fixed and `fix` broken is worse than neither.

`Lint`'s `timeout-minutes` goes 15 → 30. **Measured on this PR: the `lint` step went 1m35s → 2m24s** (job total 3m02s → 3m52s), so the coverage costs ~50s, far less than feared. The bump is insurance for the case that is not measured here — a fork PR gets no R2 sccache credentials and runs both passes stone cold — because a timeout renders as a red `Lint` indistinguishable from a real lint failure.

`.claude/rust.md` no longer tells people to run a bare `cargo clippy`.

## Making it hard to regress: `tests/lint_gate.rs`

A comment is documentation, not a gate. Two tests (two file reads, no build cost) assert the *selection* invariants, because selection is the class of failure that is silent — a lint the gate can see fails loudly; a crate the gate cannot see fails never.

- **`every_workspace_member_inherits_the_workspace_lints`** — every `workspace.members` entry has `[lints] workspace = true`, with `gen/proto` as a named exemption (generated; `[lints.clippy] all = "allow"`). Without this, `--workspace --all-targets` lints a newly added member with *stock* clippy — no `unwrap_used`, no `indexing_slicing`, no `map_err_ignore`, no `undocumented_unsafe_blocks` — and reports green. The same silent hole, one level down. The member list is read from the manifest, never copied, or it would stop covering exactly the crates it exists to catch.
- **`the_lint_script_selects_the_whole_workspace`** — both clippy lines in `devenv.nix` still carry `--workspace --all-targets`. This restates the implementation on purpose: it is the literal regression.

Both were **negative-tested** — a guard that cannot fail is worth nothing:

```
$ # dropped --workspace from one line; swapped crates/walk's [lints] for [lints.clippy]
test the_lint_script_selects_the_whole_workspace ... FAILED
test every_workspace_member_inherits_the_workspace_lints ... FAILED

`cargo clippy --all-targets --locked -- -D warnings` omits --workspace. The repo root is
itself a package, so cargo would select only the root package: ...

these workspace members do not inherit `[workspace.lints]`, so the repo's restriction lints
do not apply to them and `lint` will report green on code that violates every one of them:
["crates/walk"]
```

## Demonstration

Four cases, each injected then reverted. The branch contains none of it.

| # | Injected | Where | Old invocation | New invocation |
|---|---|---|---|---|
| A | `let deliberately_unused = 1;` | `crates/engine` `#[cfg(test)] mod tests` | **exit 0** | **exit 101** |
| B | trait impl missing a method (E0046) | `crates/engine` `#[cfg(test)] mod tests` | **exit 0** | **exit 101** |
| C | `v[0]` (`indexing_slicing`) | `crates/walk` **lib** — in the root's dep graph | **exit 101** | exit 101 |
| D | `v[0]` (`indexing_slicing`) | `crates/plugin-go` **lib** — *not* in the root's dep graph | **exit 0** | **exit 101** |

**Case B** is the headline — the exact shape that greened on master:

```rust
#[cfg(test)]
mod tests {
    trait DemoTrait { fn demo_required_method(&self) -> bool; }
    struct DemoDouble;
    impl DemoTrait for DemoDouble {}
```
```
old: cargo clippy --all-targets --locked -- -D warnings                 -> 0
new: cargo clippy --workspace --all-targets --locked -- -D warnings     -> 101
     cargo test --locked --no-run -p engine                             -> 101

error[E0046]: not all trait items implemented, missing: `demo_required_method`
error: could not compile `engine` (lib test) due to 1 previous error
```

**Case C is the control**, and it is the one that corrected the write-up: the old invocation *does* catch it, which disproves "member libs were linted with no lints" and localises the real defect to cases A, B and D.

## What the fix exposed: 323 errors across 21 crates

290 in the first workspace pass, plus 33 in `crates/plugingo-e2e` that could only surface once the first batch compiled. Nothing was suppressed to reach green.

### 194 unfulfilled `#[expect]`s (the bulk)

Nineteen crate roots carried an identical copy-pasted 14-lint `#![cfg_attr(test, expect(...))]` header. Most entries never fire — several are already handled by `clippy.toml`'s `allow-unwrap-in-tests` / `allow-panic-in-tests` — so under a gate that actually compiles test targets they are `unfulfilled_lint_expectations` errors.

Each header is **pruned to the lints that crate's tests actually trigger** — the pattern `src/lib.rs` already used, being the one file the old gate did lint. Four crates (`proc`, `plugin-nix`, `lock`, `builtins`) needed no header at all. Net −360 lines of copy-paste.

### `crates/e2e` / `crates/plugingo-e2e`

Each `tests/*.rs` is its own crate root and inherits nothing from `src/lib.rs`, so each gets a per-file `#![expect(...)]` listing only what it triggers. No `cfg_attr(test, …)`: those targets are only ever compiled as tests. Package-level `[lints.clippy]` was tried first and rejected by cargo — `cannot override 'workspace.lints' in 'lints'`.

### Real code fixes (not suppressed)

| Lint | Where | Fix |
|---|---|---|
| `items_after_test_module` | `core/hmemoizer/cell.rs` | ~107 lines of production items moved above `mod tests`. Verified byte-identical. |
| `indexing_slicing` x2 | `plugin-go/driver_lint.rs` | `merged[..prior]` to `merged.iter().take(prior)`. `prior` is snapshotted from `merged.len()` and `merged` only grows, so the two are exactly equivalent. **Production code**, covered by 5 existing tests. |
| `dropping_copy_types` x2 | `core/blocking.rs`, `plugin-buildfile/run_file.rs` | `drop(x)` on a `Copy` value was a silent no-op; now a named `_`-prefixed binding. |
| `map_err_ignore` | `e2e/tests/deps.rs` | `.map_err(\|_\| …)` to `.context(…)` — banned repo-wide, so fixed not exempted. Also preserves `Elapsed` as the error source, which the old form discarded. |
| `redundant_closure` | `engine/sandbox_cleaner.rs` | `spawn(move \|\| walk())` to `spawn(walk)` |
| `needless_borrow` x2, `iter_kv_map`, `manual_repeat_n`, `unnecessary_get_then_check` x3 | `plugin-go` | mechanical |
| `unused_imports` | `plugingo-e2e/tests/codegen.rs` | dead `PermissionsExt` import removed |
| `doc_lazy_continuation` | `plugingo-e2e/tests/common/mod.rs` | blank `///` before a paragraph absorbed into a list item |
| `type_complexity` x3 | `core/hmemoizer/{cell,mod}.rs`, `engine/result.rs` | named type aliases |

### Three new suppressions, argued individually

1. **`#[expect(clippy::await_holding_lock)]`** on one test in `core/blocking.rs`. `exclusive()` returns a guard whose whole job is to serialize entire tests; it must outlive the awaits it brackets. `#[tokio::test]` drives the body via `block_on` on the calling thread and `MutexGuard` is `!Send`, so it cannot migrate. Scoped to the one test, not the crate.
2. **`#[expect(clippy::multiple_unsafe_ops_per_block)]`** on `core/fsutil.rs::spawn_fd_holder`. `pre_exec` plus the `libc::dup` inside its closure. They cannot be split *in place*: a closure body inside an `unsafe` block inherits that context, so a nested `unsafe { libc::dup(raw) }` is `unused_unsafe`. The existing SAFETY comment covers both.
3. **`#![allow(dead_code, unused_imports, reason = …)]` kept, not converted to `expect`,** in the two shared `tests/common/mod.rs`. Each test binary includes the file textually and uses a different subset, so an `expect` would be unfulfilled in most of them. Only the missing `reason` was added.

No blanket `#[allow]` anywhere else; nothing uses `map_err(|_| …)`.

## Verification

`RUSTC_WRAPPER=""`, `aarch64-apple-darwin`:

| Check | Result |
|---|---|
| `cargo clippy --workspace --all-targets --locked -- -D warnings` | exit 0 |
| `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` | exit 0 |
| `cargo fmt --check` (all 34 quality crates) | exit 0 |
| `cargo test --locked --no-run -p core -p engine -p plugin-go -p plugin-buildfile -p e2e -p plugingo-e2e` | exit 0 |
| `cargo test -p heph --test lint_gate` | 2 passed (+ both negative-tested) |
| `cargo test -p core --lib` | 107 passed |
| `cargo test -p plugin-go --lib merge` | 7 passed — covers the `driver_lint` rewrite |
| `cargo test -p engine --lib sandbox_cleaner` | 17 passed |
| `cargo test -p plugin-buildfile --lib run_file` | 71 passed |
| `nix-instantiate --parse devenv.nix` | OK |
| **CI `Lint` on `ubuntu-latest`** | **pass** on both commits — the Linux half of the `#[expect]` pruning |
| CI `Lint` step duration | 2m24s (was 1m35s) |

`clippy --workspace` and `cargo test --no-run` are run as separate checks on purpose: a clean clippy has coexisted with a crate that fails to compile under `cargo test` twice this week, which is why this PR exists.

## Deliberately not done here — three coverage decisions for you

Each widens what CI builds, so each is a call to make rather than something to slip into a lint fix. All three are now recorded in the `devenv.nix` comment so they are not rediscovered.

1. **`Lint` runs on `ubuntu-latest` only.** macOS-only code is compiled by `Test darwin/arm64` but nothing there passes `-D warnings`, so it is clippy-linted nowhere — including the Mach `sweep_threads` block in `src/diag.rs`, the largest unsafe region in the tree. This PR makes that gap sharper: `#[expect]` is fulfilled-or-error, and the pruned headers are pinned to the lint set that fires on Linux, so a Mac-local `lint` can now disagree with CI in either direction. An audit found the only OS-gated test item today (`core/fsutil.rs:319`) triggers none of the pruned lints, so nothing is broken now. Adding a `Lint darwin/arm64` leg costs one runner (~3-6 min) and is the only thing that would ever lint the Mach code.
2. **`--no-default-features` (`fuse-sandbox` off) is linted nowhere and built nowhere.** A documented knob in two manifests that CI has never compiled. Either cover it with a cheap `cargo check --workspace --no-default-features`, or say it is unsupported and stop offering it.
3. **`expect` vs `allow` for crate-root test exemptions.** The tree already disagrees with itself: `crates/config` and `crates/selfupdate` use `#![cfg_attr(test, allow(..., reason = …))]` precisely because not every listed lint fires, while `src/lib.rs` says `expect` "keeps these honest". This PR entrenches `expect` in 15 more crates. Worth picking one shape workspace-wide and stating it once, or it gets re-litigated per crate. (`--all-targets` also excludes doctests — 10 blocks, noted, not worth a pass.)


raphaelvigee:
Rebased onto `39ac1b34` (was `be5b01fe`). The rebase itself was clean — no conflicts — but as expected it surfaced four fresh clippy errors in code that merged under the old, hole-ridden gate (#247, #249, #255, #257, #258, #261). They are in the extra commit `fix(lint): fix the four errors the rebase surfaced in newly-landed code`:

| Error | Where | Fix |
|---|---|---|
| `unused_imports` (`WalkEntryKind`) | `core/hartifactcontent/tar_index.rs`, a `#[cfg(unix)] #[test]` | dead import removed |
| `map_err_ignore` | `engine/result.rs`, `tokio::time::timeout` | `.context(…)` — banned repo-wide, so fixed not exempted; also keeps `Elapsed` as the error source |
| `await_holding_lock` | `engine/request_state.rs`, `the_exit_gate_is_held_across_the_trim_retry` | scoped `#[expect]` with a reason, mirroring the `hcore::blocking::exclusive()` one already in this branch. The guard serializes whole tests, so it is meant to outlive the awaits, and `#[tokio::test]` drives the body via `block_on` on the calling thread so the `!Send` guard cannot migrate. The three `gc.rs` users are plain `#[test]` and need nothing. |
| `manual_is_multiple_of` | `e2e/tests/query.rs` | mechanical |

No blanket `#[allow]`, no weakening of the gate. Three of the four are in test targets the old invocation never built — which is the thesis of this PR, demonstrated once more by the merges that happened while it was open.

The two coverage decisions this PR left open — a `Lint darwin/arm64` leg, and `--no-default-features` — are now closed in **#262**, stacked on this branch. The third (`expect` vs `allow` for crate-root test exemptions) is left alone deliberately; it is a convention call.

One finding from #262 worth knowing here: the obvious `cargo clippy --workspace --no-default-features` **does not turn `fuse-sandbox` off** — `e2e`/`plugingo-e2e`/`testkit` pull the root `heph` package with default features and cargo unifies across the selection. Verified by injection: exits 0 on a deliberate error in the feature-off arm. Same silent-selection class as this PR, one axis over.

PR #262, opened 2026-07-29
ci(lint): lint macOS and the `fuse-sandbox`-off arm, as a per-platform matrix

raphaelvigee:
> ## ⚠️ Read this first: you must change branch protection when this merges
>
> The lint job is a **matrix**, so its status contexts are the per-leg names. **No check named plain `Lint` will ever report again.** The ruleset currently requires exactly that string.
>
> **Required status checks — remove:**
> ```
> Lint
> ```
> **Required status checks — add:**
> ```
> Lint linux/amd64
> Lint darwin/arm64
> ```
>
> Those two strings are taken from a **real run of this matrix** ([30485344790](https://github.com/hephbuild/heph/actions/runs/30485344790)), not inferred: GitHub publishes the rendered `name:` with no parenthesised matrix suffix.  A wrong string here blocks every PR just as badly as a missing one.
>
> Consequences until you make that change:
> - a required `Lint` can no longer report, so **every open PR stays blocked**;
> - a **red lint will not block a merge**, because neither new context is required yet.
>
> Inside the repo, `release` and `cleanup` still `needs: lint`, so a red lint does stop a release on master regardless. That is the only enforcement the tree itself owns.

---

Closes the two coverage gaps #259 left open as decisions for you, and reshapes the lint job into a per-platform matrix.

## (a) macOS code was clippy-linted nowhere

`Test darwin/arm64` compiles it, but nothing there passes `-D warnings`. So the Mach `sweep_threads` block in `src/diag.rs` — the largest unsafe region in the tree — plus `crates/proc/src/proc_exec/imp_macos.rs` (609 lines) and `process_watcher/kqueue_macos.rs` (286 lines) had never been through `undocumented_unsafe_blocks`, `multiple_unsafe_ops_per_block`, `indexing_slicing` or `unwrap_used`.

The `lint` job is now a matrix over `linux/amd64` + `darwin/arm64`, both running the **same** `lint` script — no per-OS variant. The failure being closed is a Mac developer's local `lint` disagreeing with CI in either direction (`#[expect]` is fulfilled-or-error, so a lint that fires on one OS and not the other is a hard error on the other), and that only stays closed if every leg runs the same command the developer does.

`fail-fast: false` is load-bearing, not boilerplate: with fail-fast on, the first red leg cancels the other, so the platform you did not see reports `cancelled` rather than a result — hiding exactly the half you need on the one case that matters.

**It found nothing.** All three clippy passes exit 0 on `aarch64-apple-darwin`. Every unsafe block in the Mach path already carried a SAFETY comment, each holds one unsafe op, and the port array is consumed via `.iter().take(MAX_THREADS)` rather than indexed. The `#[expect]` headers #259 pruned against the Linux lint set are fulfilled on macOS too. Finding nothing is the outcome the leg exists to *keep* true.

Two stale comments in that newly-covered region are fixed here, because **no lint can catch either**: a doc block on `sweep_threads` describing a helper that no longer exists and asserting `mach2` had *not* been pulled in (it has — `Cargo.toml` adds it and line 263 calls it), and a duplicated `// SAFETY:` whose first copy documented the *next* block's invariant. `undocumented_unsafe_blocks` is satisfied by any preceding SAFETY line, so it stayed green while the text pointed at the wrong operation.

## (b) `--no-default-features` — and the obvious spelling of it is a no-op

`crates/sandboxfuse/src/stub.rs`, the whole feature-off arm, is compiled by nothing today on any platform, and must still satisfy its callers in `crates/engine` and `crates/driver-bridge`. A signature drift in `imp.rs` that `stub.rs` did not follow is green right now.

**The substance of this half: `cargo clippy --workspace --no-default-features` does not turn `fuse-sandbox` off.** `crates/e2e`, `crates/plugingo-e2e` and `crates/testkit` depend on the root `heph` package with default features on, and cargo unifies features across the whole selected set, so that edge switches it straight back on.

Verified rather than reasoned — an `indexing_slicing` error injected into `stub.rs`, the file compiled *only* when the feature is off:

| Invocation | Exit |
|---|---|
| `cargo clippy --workspace --all-targets --no-default-features` | **0** |
| `… --exclude e2e --exclude plugingo-e2e --exclude testkit …` | **101** |
| `cargo clippy -p sandboxfuse --all-targets --no-default-features` | 101 |
| `cargo clippy -p heph --all-targets --no-default-features` | 101 |

Landing the naive form would have shipped a third pass that reports green while re-linting the arm pass 1 already covered — the same silent-selection failure #259 exists to fix, one axis over.

Because that regresses invisibly (a new member with `heph = { path = "../.." }` re-enables it and the pass keeps reporting green), **`lint` asserts the feature is off before linting it**, with `cargo tree -i fuser`, rather than trusting the flag:

```
$ cargo tree --workspace --exclude e2e --exclude plugingo-e2e --exclude testkit --no-default-features -i fuser
exit 101   # fuser absent — feature genuinely off
$ cargo tree --workspace --no-default-features -i fuser
exit 0     # fuser still in the graph — the flag is a no-op
```

Absence is matched on `did not match any packages`, not on the exit code alone: reading *any* cargo failure as "feature off" would wave the wrong arm through. A positive control asserts `fuser` **is** present with the feature on, so "absent because off" cannot be confused with "absent because gone". `fix` carries the matching pass with the matching exclusions.

The feature-off arm compiles and lints clean. No code changes were needed.

## Guards: `tests/lint_gate.rs` goes 2 tests → 9

The selection has five axes and each has already gone wrong or been left open: which packages (`--workspace`), which targets (`--all-targets`), which feature set, which platform, and **whether a red leg has any consequence at all**.

That last one is the new silent-pass class, and it is the one worth reading: a leg can be made to conclude `success` with clippy never running. `continue-on-error: true` converts a failure into a success; an `if:` on the Lint step skips it entirely. Either turns the leg's own check green over an unlinted platform, and `needs:` treats it as satisfied. Both are now asserted absent (the `if:` check scoped to the Lint step, since `sccache stats` legitimately has `if: always()`).

**23 breaks injected across the round, each caught.** The 10 for the current shape:

| Break | Caught by |
|---|---|
| matrix loses its darwin leg | `the_lint_gate_runs_on_macos_as_well_as_linux` |
| matrix loses its linux leg | same |
| leg runs something narrower than `lint` | same |
| `strategy:` removed (silently one platform) | same |
| `fail-fast: true` | `the_lint_job_is_a_matrix_over_both_supported_oses` |
| duplicated os/arch coordinate | same |
| `continue-on-error:` on the leg | `a_lint_leg_cannot_report_success_without_running_the_lint` |
| `if:` on the Lint step | same |
| `release` stops depending on lint | `lint_gates_release_and_cleanup` |
| `cleanup` stops depending on lint | same |

plus, from the `--no-default-features` and fmt axes: drop `--workspace` · delete the ndf pass · delete the `cargo tree` probe · drift the probe from the pass · drop an `--exclude` · add an extra `--exclude engine` · `fix` loses its default pass · `fix` loses its ndf pass · drop a crate from `qualityCrates` · swap a member's `[lints]` for `[lints.clippy]`.

The exclusion list is **derived from the member manifests, never copied**, and checked for **equality** rather than containment — an *extra* `--exclude engine` or `--exclude sandboxfuse` guts the pass just as thoroughly and does not put `fuser` back in the graph, so neither the probe nor a subset check can see it.

`assert_eq!(clippy_lines.len(), 2)` is relaxed to `>= 3`: it asserted an exact count, so it failed when coverage was *added* — the wrong direction for a coverage guard.

Two negative tests initially proved nothing and were redone, recorded because a negative test that silently edits the wrong thing is indistinguishable from a guard that works: the first `- os: darwin` block in the workflow belongs to `build:`, so an unscoped rewrite deleted the wrong job's leg. Every workflow break is now applied inside the `lint` job only, via a helper that asserts the pattern was found there.

## What `needs: lint` actually guarantees

`needs:` on a matrix job waits for **every** leg and is satisfied only if all succeeded. Confirmed against run 30485344790 rather than trusting the docs: `bin_e2e` (`needs: [gen, build, …]`) started at 20:24:50 — four seconds after the *last* `build` leg finished (20:24:46), not after the first (19:55:52). So `release`/`cleanup` gate on the whole matrix with no aggregator, and their `needs` entries required no edit after the rename.

## Cost

Measured on the linux leg (run 30477623663, warm):

| Pass | Time |
|---|---|
| `clippy --workspace --all-targets` | 2m02s |
| `… --all-features` | **0.26s** |
| `… --no-default-features` | **13.70s** |
| **`Lint` step total** | **2m18s** (was 2m07s) |

**+11s** for the feature-off coverage; it rebuilds exactly the four packages the feature reaches (`sandboxfuse`, `driver-bridge`, `engine`, `heph`). The 0.26s on `--all-features` is the measurement behind a correction to #259's claim: that pass is ~a no-op even *with* `--workspace`, because `--workspace` already unifies every member feature on. It is kept as cheap insurance and the comment now says so.

The darwin leg measured **9m49s / 10m24s** job total (Nix setup ~3m15s, clippy ~6m). That is the **cold** number and should not be read as steady state: sccache does not cache `clippy-driver` at all (~230 non-cacheable calls, 0 non-cacheable compilations), and `setup-nix` saves the cargo registry cache only on `master` — PRs are restore-only by design, so `cargo-aarch64-apple-darwin-lint-*` will not exist until this lands.

**The real price is macOS runner contention, not minutes.** The repo has a hard 5-concurrent-macOS cap and this adds a fourth macOS job to the `needs: gen` wave. Measured against baseline run 30466726477: `Build darwin/arm64` — the darwin critical path — was **not** delayed, and the workflow critical path (`Test linux/amd64`) did not move. But `Warm Binary E2E darwin/arm64` took the queue instead, and its margin over `Binary E2E darwin/arm64` collapsed from **803s to 4s**. It is normally not the binding constraint, so the delay is largely absorbed today — but that is a property of the current numbers, not a guarantee. Two attempts to re-measure this precisely under the matrix shape were abandoned when the runner pool saturated with eight sibling branches in flight (all four macOS jobs still queued 7.5 min in); CI on this head is the evidence rather than a synthetic campaign.

## Still open — recorded in `devenv.nix` rather than left to be rediscovered

- Doctests: `--all-targets` excludes them (10 blocks).
- `cargo test --no-default-features` is not run — the arm is compiled and linted, never *executed*. `sandboxfuse`'s `fuse_sandbox_is_a_default_feature` asserts the feature **is** on, so a test leg needs that `#[cfg]`-split first.
- **The arch axis is confounded.** `aarch64` is only ever linted together with macOS, and there is no `Lint linux/arm64`. `Test linux/arm64` cannot stand in: plain rustc reports `unfulfilled_lint_expectations` for its *own* lints but silently ignores an unfulfilled expectation on a **tool** lint it does not know, so a `target_arch`-conditional `#[expect(clippy::…)]` would be invisible in every job that exists. Per-platform behaviour is your call, so it is written down, not decided here. (Cross-clippy from Linux is not an option: `ring`/`aws-lc-sys` compile C for the target, `macfuse-stubs` is Darwin-only in `devenv.nix`, and `--all-targets` needs a real Darwin linker.)
- `expect` vs `allow` for crate-root test exemptions is left alone — the tree genuinely contradicts itself (`config`/`selfupdate` use `allow`, `src/lib.rs` argues for `expect`) and picking a side is a convention call.

## Review

`code-quality` and `feature-quality` were consulted at design and on each finished shape. Findings implemented rather than noted, including two that were genuine defects in my own new tests: a fmt-coverage guard that was **red as written** (`-p xstarlark-fmt ` with a trailing space, and it is the last entry in `qualityCrates`), and a probe-existence guard that matched its own explanatory comment. One `feature-quality` design recommendation was **overruled with evidence** — it proposed `cargo clippy --workspace --all-targets --no-default-features` as clean; the injection test above shows that invocation exits 0 on a deliberate error in the feature-off arm.

### DIFF
--- a/.github/workflows/heph.yml
+++ b/.github/workflows/heph.yml
@@ -427,10 +427,14 @@ jobs:
     name: Lint
     runs-on: ubuntu-latest
     # The gate now check-builds every workspace member's libs *and* test targets,
-    # twice (default features, then --all-features), where it used to build the
-    # root package alone in ~1m35s. A fork PR gets no R2 sccache credentials and
-    # runs both passes stone cold. A timeout renders as a red `Lint`
-    # indistinguishable from a real lint failure, so the headroom is deliberate.
+    # three times (default features, --all-features, --no-default-features),
+    # where it used to build the root package alone in ~1m35s. A fork PR gets no
+    # R2 sccache credentials and runs every pass stone cold — and clippy's work
+    # is not sccache-cacheable at all (`sccache` does not cache `clippy-driver`,
+    # which is why the Lint job's own stats report ~200 "non-cacheable calls"),
+    # so a warm cache only saves the dependency compiles. A timeout renders as a
+    # red `Lint` indistinguishable from a real lint failure, so the headroom is
+    # deliberate.
     timeout-minutes: 30
     needs: gen
     steps:
@@ -447,6 +451,64 @@ jobs:
           # Distinct from the linux/amd64 build job, which uses the same target
           # triple: sharing a cargo cache key means whichever job saves first
           # freezes the entry and the other's crate downloads never persist.
+          cache-suffix: "-lint"
+          r2-secret: ${{ secrets.CF_SSCACHE }}
+
+      - name: Lint
+        shell: devenv shell bash -- -e {0}
+        run: lint
+
+      - name: sccache stats
+        if: always()
+        shell: devenv shell bash -- -e {0}
+        run: sccache --show-stats
+
+  # macOS-only code was clippy-linted nowhere. `Test darwin/arm64` compiles it,
+  # but nothing there passes `-D warnings`, so the Mach `sweep_threads` block in
+  # `src/diag.rs` — the largest unsafe region in the tree — plus
+  # `crates/proc/src/proc_exec/imp_macos.rs` and `process_watcher/kqueue_macos.rs`
+  # had never been through `undocumented_unsafe_blocks`,
+  # `multiple_unsafe_ops_per_block`, `indexing_slicing` or `unwrap_used`.
+  #
+  # It runs the *same* `lint` script as the Linux leg rather than a trimmed one.
+  # The failure this closes is a Mac developer's local `lint` disagreeing with
+  # CI in either direction — `#[expect]` is fulfilled-or-error, so a lint that
+  # fires on one OS and not the other is a hard error on the other — and that
+  # only stays closed if both run the same command. A darwin-only variant would
+  # recreate the divergence it exists to remove.
+  #
+  # A separate job, deliberately not a `matrix` on `lint`: the branch ruleset
+  # requires a status context named literally `Lint`. A matrix would publish
+  # `Lint linux/amd64` / `Lint darwin/arm64` and never `Lint` again, so every
+  # open PR would block on a required check that can no longer report, and the
+  # workflow edit and the ruleset edit cannot land atomically.
+  #
+  # Cost, measured rather than assumed — see the PR body. It is a fourth
+  # concurrent `macos-latest` job in the `needs: gen` wave, which is the real
+  # price: the darwin build chain has been observed on the critical path, and
+  # macOS jobs in this repo have queued for up to ~2m30s waiting for a slot.
+  lint_darwin:
+    name: Lint darwin/arm64
+    runs-on: macos-latest
+    # Same reasoning as `lint` above, and macOS starts colder: the Nix setup
+    # alone is measured at 3m44s-4m42s on this runner.
+    timeout-minutes: 30
+    needs: gen
+    steps:
+      - name: Download repo
+        uses: actions/download-artifact@v7
+        with:
+          name: repo
+          path: .
+
+      - name: Setup Nix + devenv
+        uses: ./.github/actions/setup-nix
+        with:
+          cargo-target: aarch64-apple-darwin
+          # Distinct from `Test darwin/arm64`, which uses the same target
+          # triple: actions/cache skips the save when the key already exists, so
+          # two jobs sharing a key means whichever saves first freezes the entry
+          # and the other's crate downloads never persist.
           cache-suffix: "-lint"
           r2-secret: ${{ secrets.CF_SSCACHE }}
 
@@ -653,7 +715,7 @@ jobs:
 
   release:
     name: Release
-    needs: [gen, upload_artifacts, lint, test, bin_e2e]
+    needs: [gen, upload_artifacts, lint, lint_darwin, test, bin_e2e]
     if: github.ref == 'refs/heads/master'
     runs-on: ubuntu-latest
     timeout-minutes: 10
@@ -705,7 +767,7 @@ jobs:
 
   cleanup:
     name: Cleanup
-    needs: [gen, build, lint, test, bin_e2e, upload_artifacts]
+    needs: [gen, build, lint, lint_darwin, test, bin_e2e, upload_artifacts]
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
--- a/.github/workflows/heph.yml
+++ b/.github/workflows/heph.yml
@@ -423,20 +423,56 @@ jobs:
       - name: Check plugin ABI version
         run: scripts/abi-check.sh "origin/${{ github.base_ref }}"
 
-  lint:
-    name: Lint
-    runs-on: ubuntu-latest
-    # The gate now check-builds every workspace member's libs *and* test targets,
-    # three times (default features, --all-features, --no-default-features),
-    # where it used to build the root package alone in ~1m35s. A fork PR gets no
-    # R2 sccache credentials and runs every pass stone cold — and clippy's work
-    # is not sccache-cacheable at all (`sccache` does not cache `clippy-driver`,
-    # which is why the Lint job's own stats report ~200 "non-cacheable calls"),
-    # so a warm cache only saves the dependency compiles. A timeout renders as a
-    # red `Lint` indistinguishable from a real lint failure, so the headroom is
-    # deliberate.
+  # The gate check-builds every workspace member's libs *and* test targets,
+  # three times (default features, --all-features, --no-default-features),
+  # where it used to build the root package alone in ~1m35s. A fork PR gets no
+  # R2 sccache credentials and runs every pass stone cold — and clippy's work is
+  # not sccache-cacheable at all (`sccache` does not cache `clippy-driver`,
+  # which is why these jobs' own stats report ~200 "non-cacheable calls"), so a
+  # warm cache only saves the dependency compiles. A timeout renders as a red
+  # lint leg indistinguishable from a real lint failure, so the 30min headroom
+  # is deliberate.
+  #
+  # Both platforms, because macOS-only code was clippy-linted nowhere.
+  # `Test darwin/arm64` compiles it, but nothing there passes `-D warnings`, so
+  # the Mach `sweep_threads` block in `src/diag.rs` — the largest unsafe region
+  # in the tree — plus `crates/proc/src/proc_exec/imp_macos.rs` and
+  # `process_watcher/kqueue_macos.rs` had never been through
+  # `undocumented_unsafe_blocks`, `multiple_unsafe_ops_per_block`,
+  # `indexing_slicing` or `unwrap_used`.
+  #
+  # Every leg runs the *same* `lint` script — no per-OS variant. The failure
+  # being closed is a Mac developer's local `lint` disagreeing with CI in either
+  # direction (`#[expect]` is fulfilled-or-error, so a lint that fires on one OS
+  # and not the other is a hard error on the other), and that only stays closed
+  # if every leg runs the same command as the developer does.
+  lint_matrix:
+    name: Lint ${{ matrix.os }}/${{ matrix.arch }}
+    runs-on: ${{ matrix.runs-on }}
     timeout-minutes: 30
     needs: gen
+    strategy:
+      # Both legs always report. With fail-fast the second is cancelled the
+      # moment the first goes red, and `cancelled` is exactly the state that
+      # tells you nothing about the platform you did not see.
+      fail-fast: false
+      matrix:
+        include:
+          - os: linux
+            arch: amd64
+            runs-on: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+
+          # macos-latest is Apple Silicon. There is no `Lint linux/arm64` leg:
+          # aarch64 is covered here on the OS axis, and an `#[expect]` that is
+          # unfulfilled only on aarch64-linux would be invisible anyway — plain
+          # rustc silently ignores an unfulfilled expectation on a *tool* lint
+          # it does not know, so `Test linux/arm64` cannot stand in for one.
+          # See the note in devenv.nix.
+          - os: darwin
+            arch: arm64
+            runs-on: macos-latest
+            target: aarch64-apple-darwin
     steps:
       - name: Download repo
         uses: actions/download-artifact@v7
@@ -447,9 +483,10 @@ jobs:
       - name: Setup Nix + devenv
         uses: ./.github/actions/setup-nix
         with:
-          cargo-target: x86_64-unknown-linux-gnu
-          # Distinct from the linux/amd64 build job, which uses the same target
-          # triple: sharing a cargo cache key means whichever job saves first
+          cargo-target: ${{ matrix.target }}
+          # Distinct from the build and test jobs, which use the same target
+          # triples: actions/cache skips the save when the exact key already
+          # exists, so two jobs sharing a key means whichever saves first
           # freezes the entry and the other's crate downloads never persist.
           cache-suffix: "-lint"
           r2-secret: ${{ secrets.CF_SSCACHE }}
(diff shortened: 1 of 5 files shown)


raphaelvigee:
## Measured CI cost — first run, `30477623663`

Both new legs are **green on their first run**. No macOS code needed fixing, and the feature-off arm compiles and lints clean.

### (b) `--no-default-features`, on the existing linux `Lint` job

| Pass | This run | PR #259's run (`30477488524`, same hour, warm) |
|---|---|---|
| `clippy --workspace --all-targets` | 2m02s | — |
| `clippy … --all-features` | **0.26s** | — |
| `clippy … --no-default-features` | **13.70s** | *(did not exist)* |
| **`Lint` step total** | **2m18s** | **2m07s** |

**+11s.** The third pass rebuilds exactly the four packages the feature reaches — `sandboxfuse` (the stub arm), `driver-bridge`, `engine`, `heph` — and nothing else, because the excludes keep `e2e`/`plugingo-e2e`/`testkit` out of the selection.

The 0.26s on `--all-features` is the measurement behind the correction in the commit body: that pass is still ~a no-op *with* `--workspace`.

### (a) `Lint darwin/arm64`

| Step | Time |
|---|---|
| Setup Nix + devenv | 3m15s |
| `clippy --workspace --all-targets` | 5m57s |
| `clippy … --all-features` | 1.06s |
| `clippy … --no-default-features` | 18.21s |
| `cargo fmt --check` | ~2s |
| **Job total** | **9m49s** |

This is the **cold** number and should not be read as steady state: the `cargo-aarch64-apple-darwin-lint-*` registry cache did not exist yet, and sccache's Rust hit rate was **9.29%** (1127 compile requests, 430 hits, 229 non-cacheable calls, 0 non-cacheable compilations). C/C++ and assembler hit 100% — it is the Rust units that were cold. The linux `Lint` sits at ~90% on a warm key.

### The honest cost: macOS runner concurrency, not minutes

The repo gets **3 concurrent macOS runners**. Adding a 4th job pushes one into the queue. Measured against baseline run `30466726477` (same workflow, no darwin lint leg), delay between job *creation* and *start*:

| Job | Baseline | This run |
|---|---|---|
| `Build darwin/arm64` | 4s | 12s |
| `Lint darwin/arm64` | — | 11s |
| `Test darwin/arm64` | 4s | **2m49s** |
| `Warm Binary E2E darwin/arm64` | 4s | **9m19s** |

`Build darwin/arm64` — the job on the darwin critical path — was **not** delayed. `Warm Binary E2E darwin/arm64` took the queue instead, and it is normally not the binding constraint (baseline: it finished at 15:45:03, `Build darwin/arm64` at 15:53:43, and `Binary E2E darwin/arm64` needs both). So the delay is largely absorbed today.

**That is a property of the current numbers, not a guarantee.** If `Build darwin/arm64` ever gets faster, `bin_e2e_warm` becomes the constraint and this leg starts costing real wall-clock. Worth revisiting then — the fallback is to fold the darwin clippy pass into `Test darwin/arm64` as a step, which costs zero runner slots and reuses its already-paid 3m44s Nix setup, at the price of a lint failure reddening a check named `Test`.

raphaelvigee:
## Review round: one of my own new tests was red, and three axes were unguarded

`feature-quality` returned **BLOCKED** on the finished diff and was right. Pushed `b930606c`.

### The red test

`the_fmt_pass_covers_every_workspace_member` — added in the previous commit to guard the `cargo fmt --check` selection — matched `-p {name} ` with a mandatory trailing space. `xstarlark-fmt` is the **last** entry in `qualityCrates`, so the file contains `-p xstarlark-fmt"` and never `-p xstarlark-fmt `. It would have gone red on all three test legs on the first push, with a message that was a lie ("xstarlark-fmt is not in qualityCrates" — it is), inviting a fix-by-duplicate.

It also searched the whole file rather than the `qualityCrates` value, and `scripts.tst.exec` carries `-p plugin-stabby` and `-p plugin-sdk` too — so deleting either from `qualityCrates` left the test green while `cargo fmt --check` stopped looking at the crate. Exactly the hole it was written to close.

### Three axes with no guard at all

| Axis | Break that was green | Now |
|---|---|---|
| the `cargo tree -i fuser` probe's own existence | delete it — the previous commit's whole argument rests on it | asserted, **and** asserted to carry the same `--exclude` set as the pass it guards |
| exclusion-list **equality** | `--exclude engine` / `--exclude sandboxfuse` — each guts the pass, and neither puts `fuser` back in the graph, so the runtime probe can't see it either | set equality, not containment |
| `lint_darwin` in `release.needs`/`cleanup.needs` | remove them and the leg can go red forever with nothing blocked (it is not in the ruleset, so these are its only teeth) | asserted for every job id starting with `lint`, tokenised (`lint` is a prefix of `lint_darwin`) |

Plus `fix_selects_the_same_code_as_lint` did not cover its own name: it compared only the feature-off axis, so deleting `fix`'s **default** pass satisfied every assertion via the surviving line.

### Positive control on the probe

`cargo tree -i fuser` failing means "absent" — but absent-because-off and absent-because-gone are indistinguishable from one side. If `fuser` ever leaves the graph for an unrelated reason the check would pass forever while proving nothing. A second sub-second call asserts `fuser` **is** present with the feature on. The failure branch also now separates absence (`did not match any packages`) from any other cargo error — an ambiguous `fuser` after a transitive version bump would otherwise have read as "feature off".

### Verification without a build

This machine's volume hit 100% and `target/` could not be rebuilt. These tests read files and do string work with **zero crate dependencies**, so `tests/lint_gate.rs` was compiled standalone with `rustc --edition 2024` and every test fn run directly. All six pass. Thirteen breaks were then injected one at a time and **each was caught**:

drop `--workspace` · delete the ndf pass · delete the probe · drift the probe from the pass · drop an `--exclude` · add an extra `--exclude engine` · `fix` loses its default pass · `fix` loses its ndf pass · drop a crate from `qualityCrates` · move the darwin job to ubuntu · point the darwin job at something other than `lint` · drop `lint_darwin` from `release.needs` · swap a member's `[lints]` for `[lints.clippy]`

`rustfmt --check` and `nix-instantiate --parse` clean. CI runs the real `cargo test`.

---

## One measured consequence you should decide on

Adding the 4th concurrent macOS job did **not** delay `Build darwin/arm64` (the darwin critical path) and did **not** move the workflow critical path, which is still `Test linux/amd64`. But it did push `Warm Binary E2E darwin/arm64` to the back of a hard **5-concurrent-macOS-runner** cap, and that job's safety margin collapsed:

| | `Warm Binary E2E darwin` ends | `Binary E2E darwin` starts | margin |
|---|---|---|---|
| baseline `30466726477` | 15:45:03 | 15:58:26 | **803s** |
| this branch `30477623663` | 18:12:19 | 18:12:23 | **4s** |

The warm job exists so `Binary E2E darwin/arm64` does not start against a cold sccache; `heph.yml` states it "is long finished before `bin_e2e` starts". It still was, by 4 seconds. The next slightly slower runner inverts it.

The mechanism is proven, not inferred — two consecutive 3-4s slot handoffs between the two concurrent runs:

```
run 259 Warm Binary E2E darwin ended 18:01:11 -> run 262 Test darwin    started 18:01:15
run 259 Test darwin/arm64      ended 18:07:42 -> run 262 Warm E2E darwin started 18:07:45
```

Caveat on attribution: the queueing came from two runs overlapping (8 macOS demands against 5 slots). A run in isolation places 4 ≤ 5 and queues nothing. But overlap is the normal condition here.

**This is a decision, not a defect, and it is yours.** Options: (a) accept it and record the margin in the `bin_e2e_warm` comment; (b) gate `bin_e2e_warm` for darwin on `github.ref == 'refs/heads/master'`, returning per-run macOS demand to 3 on PRs; (c) fold the darwin clippy into `Test darwin/arm64` as a step — zero new runner slots, but a lint failure then renders as a red `Test darwin/arm64`. I have not changed `bin_e2e_warm`, because narrowing when it runs is a coverage decision rather than a lint one.

raphaelvigee:
## CI green, second run — `b930606c`

All 21 checks pass, 0 failures. Both lint legs green.

| | run 1 (`92980f29`) | run 2 (`b930606c`) |
|---|---|---|
| `Lint` step (linux, 3 passes) | 2m18s | 2m34s |
| `Lint` job total | 4m07s | 4m10s |
| `Lint darwin/arm64` step | 6m20s | 5m51s |
| `Lint darwin/arm64` job total | 9m49s | 10m24s |

The darwin leg did **not** get materially cheaper on the second run. That is expected and worth writing down: `setup-nix` saves the cargo registry cache only on `master` (PRs are restore-only, by design — the 10 GB Actions quota was LRU-evicting master's copies), so `cargo-aarch64-apple-darwin-lint-*` will not exist until this lands on master. **Treat ~10 min as the PR-side cost and expect it to fall once the key is populated from master.** Clippy's own work is never sccache-cacheable regardless (`clippy-driver` is not cached — the job reports ~230 non-cacheable calls, 0 non-cacheable compilations), so it will not approach the linux leg's 4 min.

### Summary of the two runs

**gap (a) — `Lint darwin/arm64`: landed, green on first run, found no lints.** Every unsafe block in the Mach `sweep_threads` path already carried a SAFETY comment, each holds exactly one unsafe op, and the port array is consumed via `.iter().take(MAX_THREADS)` rather than indexed. The `#[expect]` headers #259 pruned against the Linux lint set are fulfilled on macOS too. Finding nothing is the outcome the leg exists to *keep* true.

**gap (b) — `--no-default-features`: landed, green, +13.7s.** It compiles and lints `crates/sandboxfuse/src/stub.rs` and its two callers, which nothing on any platform compiled before. It needed no code fixes — but it did need the right invocation, and the obvious one is a silent no-op.

raphaelvigee:
## Reshaped: the lint legs are now a matrix, behind an aggregator

Per your decision. Pushed `a52a0aa2`.

```
lint_matrix   Lint linux/amd64     ubuntu-latest    ┐
              Lint darwin/arm64    macos-latest     ┘ fail-fast: false
                                                     │
lint          Lint                 ubuntu-latest    ◄─┘  needs: lint_matrix
                                                          if: always()
                                                          fails unless result == success
```

### Branch protection: **nothing to change** — and that is the point

A bare `strategy.matrix` on the existing job renames the checks to `Lint (ubuntu-latest)` / `Lint (macos-latest)`. The ruleset requires a context named literally `Lint`, so nothing would ever report it again and **every open PR would block forever on a required check that cannot appear** — and the workflow edit and the ruleset edit cannot land atomically, so no ordering avoids it.

The aggregator keeps that name. Confirmed by reading the wiring back through a YAML parser rather than asserting it:

```
lint.name       = "Lint"          # required context, unchanged
lint.needs      = "lint_matrix"
lint.if         = "always()"
lint_matrix.name = "Lint ${{ matrix.os }}/${{ matrix.arch }}"   # -> Lint linux/amd64, Lint darwin/arm64
lint_matrix.strategy.fail-fast = false
release.needs   = [gen, upload_artifacts, lint, test, bin_e2e]
cleanup.needs   = [gen, build, lint, test, bin_e2e, upload_artifacts]
```

The two per-platform contexts are **new and not required**, so they appear on PRs with their own logs and re-run buttons without blocking anything. If you later want them required, that is an additive ruleset change with no window where a required check cannot report.

This also supersedes the earlier note asking you to add `Lint darwin/arm64` to the ruleset — no longer needed, the aggregator covers it.

### The aggregator is the new silent-pass risk, so it is the new guard

Two wirings that both look right in review:

- **no `if: always()`** — a failed leg *skips* the aggregator, and a skipped required check is not a red one. The PR becomes un-mergeable rather than reported as failing.
- **`if: always()` and nothing else** — the job runs, its own steps pass, and it reports **green over a failed leg**. `needs` decides ordering, not outcome. That is precisely the silent-pass class this PR exists to eliminate, reintroduced at the top of it.

So it has both, and the result check is explicit. `needs.lint_matrix.result` is the aggregate over every leg, so `!= "success"` also fails closed on `cancelled` and `skipped` — deliberate: a leg that never ran means a platform that was never linted. `fail-fast: false` stops the second leg being cancelled into exactly that state.

**Executed, not just asserted.** The lint_gate tests read YAML and can only check the step's *shape*, so the shape was run. The body is extracted from the workflow with a YAML parser (it cannot drift from what CI executes) and driven with every value GitHub can produce:

| `needs.lint_matrix.result` | exit |
|---|---|
| `success` | **0** |
| `failure` | 1 |
| `cancelled` | 1 |
| `skipped` | 1 |
| *(empty)* | 1 |

### Guards: 6 → 8, extended not loosened

`the_lint_gate_runs_on_macos_as_well_as_linux` now derives the legs from which jobs actually `run: lint` and reads the concrete runner labels out of `matrix.include` (skipping `${{ matrix.runs-on }}`), so it survives the reshape rather than being relaxed to fit it. Two new tests cover the aggregator and the downstream gating.

**12 breaks injected, each caught:**

| Break | Caught by |
|---|---|
| `lint_matrix` loses its darwin leg | `the_lint_gate_runs_on_macos_as_well_as_linux` |
| `lint_matrix` loses its linux leg | same |
| a leg runs something narrower than `lint` | same |
| aggregator renamed | `the_lint_aggregator_keeps_the_required_context_and_can_fail` |
| aggregator becomes a matrix itself | same |
| `if: always()` dropped | same |
| result check dropped | same |
| `exit 1` dropped | same |
| aggregator stops needing the matrix | same |
| a third leg added the aggregator never checks | same |
| `release` stops gating on lint | `lint_gates_release_and_cleanup` |
| `cleanup` stops gating on lint | same |

The per-dependency `needs.<dep>.result` check is what makes the third-leg case fail here rather than in production.

> One negative test initially proved nothing and was redone: the first `- os: darwin` block in the file belongs to `build`, so my edit deleted the wrong job's leg and the guard was never exercised. Recorded because a negative test that silently edits the wrong thing is indistinguishable from a guard that works.

Everything else intact: the `cargo tree -i fuser` assertion and its positive control, the exclusion-list equality guard, and the measured cost notes.

Cost is being re-measured on this run — in particular whether the matrix changes the `Warm Binary E2E darwin/arm64` margin I reported collapsing from 803s to 4s. Per-run macOS demand is unchanged at 4, so I expect no change, but I will post timestamps rather than reasoning.

raphaelvigee:
## Unblocked: #259 merged mid-flight and this PR went conflicting

Worth recording, because the symptom was **no CI run at all** rather than a red one, and that reads like a stuck queue.

**What happened.** #259 was squash-merged into master as `604035f2` at 19:15:31Z, while I was pushing the matrix reshape. GitHub retargeted this PR from `raphaelvigee/lint-covers-workspace` to `master` on its own, but the branch still carried #259's individual commits — which have no counterpart in a squashed trunk. The PR went `CONFLICTING`/`DIRTY`, GitHub could not compute the merge ref, and so **no `pull_request` workflow run was ever created** for `a52a0aa2`:

```
$ gh api repos/hephbuild/heph/actions/runs?head_sha=a52a0aa2...  -> total_count: 0
$ gh api repos/hephbuild/heph/commits/a52a0aa2.../check-runs      -> total_count: 0
```

Zero checks presents identically to a healthy PR waiting in a busy queue. This is exactly the case `CLAUDE.md` warns about under stacked PRs ("expect that rebase to conflict").

**Resolution.** `git rebase --onto origin/master 69ec4668` — replaying only this PR's three commits and dropping the ones the squash absorbed.

The resolution is verified, not assumed. The rule for a stack conflict is to diff each conflicted file against the lower branch and confirm every remaining difference is deliberately yours, because a resolution can compile, pass, and still revert the change below it. Here the check is stronger than usual and came out clean:

```
$ git diff --stat 69ec4668 origin/master
(empty)
```

The tree at my old base was **byte-identical** to the squashed master, so the rebase had nothing to reconcile and could not have dropped any of #259. Confirmed from the other side too — the diff against master is now exactly this PR's own change and nothing else:

| File | |
|---|---|
| `.github/workflows/heph.yml` | +106 — matrix + aggregator |
| `devenv.nix` | +116/−56 — the `--no-default-features` pass and its `cargo tree` probe |
| `tests/lint_gate.rs` | +625 — 2 tests → 8 |
| `crates/engine/.../request_state.rs` | 1 line — the `#[expect]` reason |
| `src/diag.rs` | −7 — the stale/duplicated SAFETY comments |

#259's own content is still in place in the master-derived files (the four `--workspace --all-targets` clippy lines in `devenv.nix`, the pruned `cfg_attr` header in `src/lib.rs`).

Re-verified after the rebase: all 8 `lint_gate` tests pass, and the aggregator's extracted step body still exits 0 only on `success` and 1 on `failure`/`cancelled`/`skipped`/empty.

PR is `MERGEABLE` again at `fa80bd84` and CI is running. **This is now a standalone PR against master, not a stack.**

PR #274, opened 2026-07-29
feat(ci): enforce perfbench for real, monitor master, file issues

raphaelvigee:
## Summary

Follow-up to #270, now that `heph-bench` is actually published (that PR's bootstrap gap — every merge-base predating the harness — is closed).

- **Enforced for real**: `Compare and report` now exits non-zero when any scenario regresses past its threshold. Each individual `compare` call still uses `--allow-regression` (so one regression doesn't cut the rendered table short — every scenario/tier is always attempted and reported), but the step greps that table for `REGRESSION` and fails for real if anything shows up. No more `continue-on-error` masking it as a yellow warning. The override is unchanged from the original design: a repo admin bypassing a required check on merge — no new plumbing needed for that.
- **Runs on `push` to `master` too**, not just `pull_request`. Baseline there is `github.event.before` (the commit immediately preceding this push) rather than a merge-base computation — robust to force-pushes in a way `HEAD~1` isn't, and treated the same as "no baseline found" when it's the all-zero SHA a ref's first push reports.
- **Auto-files a GitHub issue on regression**, but only on `push`: a PR failing the check already blocks the merge, so there's nothing further to escalate there. On `master` the merge already happened — a red commit check is easy to miss — so a regression instead opens an issue (labeled `perf-regression`, created idempotently via `gh label create --force`), deduped against any already-open issue with the same title so a re-run/retry doesn't spam.

## Fixed along the way

Giving the job a `permissions:` block (needed for `issues: write`) makes that block **exhaustive** — everything not listed drops to `none`. Added `contents: read` (checkout) and `actions: read` (the `gh api .../actions/runs` lookup in "Resolve baseline release") alongside `issues: write`, or both of those would have silently broken.

## What's still manual

Making "Perf" an actual **required** status check in branch protection is a repo-settings change outside this diff — I haven't touched branch protection. Happy to do it via `gh api` if you want, but flagging it as a separate, deliberate action rather than bundling it silently into this PR.

## Test plan

- [x] YAML parses; `actionlint` clean (no new findings — pre-existing findings are the unrelated `background`/`wait-all` govet-job steps this PR doesn't touch).
- [ ] Not yet validated by a live `push`-triggered run (the next merge to `master` is what exercises that path for the first time) or by an actual detected regression (nothing here has intentionally regressed to prove the fail-for-real / issue-filing path fires) — flagging both as real gaps in coverage rather than claiming more confidence than the testing supports.

https://claude.ai/code/session_01F9mbNMBQVSrsJojDYnQcDV

### DIFF
--- a/.github/workflows/heph.yml
+++ b/.github/workflows/heph.yml
@@ -996,6 +996,7 @@ jobs:
         run: |
           BENCH=./candidate-dist/heph-bench
           any_regression=false
+          mkdir -p verdicts
           report=$(
             echo "# Perf: ${{ matrix.os }}/${{ matrix.arch }}"
             if [ "${{ steps.baseline_fetch.outputs.ok }}" != "true" ]; then
@@ -1015,6 +1016,7 @@ jobs:
                   if [ -f "$base" ] && [ -f "$cand" ]; then
                     "$BENCH" compare --baseline "$base" --candidate "$cand" \
                       --thresholds "$GITHUB_WORKSPACE/.github/perfbench-thresholds.json" \
+                      --json "verdicts/${tier}_${scenario}.json" \
                       --allow-regression
                   else
                     echo "_${scenario}: no results (run step failed — see logs)_"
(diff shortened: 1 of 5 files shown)


raphaelvigee:
Pushed a follow-up: the filed issue now includes the actual rendered regression table (`compare`'s markdown output, captured to `perf-report.md` alongside the step summary) ahead of the job/commit links, instead of just linking out to go find the numbers.

raphaelvigee:
Two more pushes, both driven by the first real CI run's actual output:

**PostHog metrics** (`39ee5109`): \`heph-bench compare\` gains \`--json <path>\` (structured per-scenario verdicts, no new dependency). A new "Push metrics to PostHog" step batches them to PostHog's \`/batch/\` capture endpoint (EU cloud, project 236603), tagged with os/arch/sha/base_sha/event_type/run_id/run_url. Runs on both \`pull_request\` and \`push\` — \`event_type\` rides along as a property so a dashboard can filter to push-only for a "master over time" trend without the workflow needing an opinion on that. Best-effort: a PostHog outage logs and moves on. **Needs the \`POSTHOG_CI_KEY\` repo secret added before this actually sends anything** — without it the step just fails to auth and moves on (never blocks the job).

**Real bugs from the first enforced-gate run** (\`52a88905\`):
- Duplicate table headers — the job called \`compare\` once per scenario instead of once per tier, so the rendered table repeated 3x. Now merges each side's 3 scenario files into one before comparing.
- Tier B crashed outright (\`spawn heph r build //go/...: No such file or directory\`) — \`Dist::locate\` stored the \`--dist\` path verbatim, and Tier B spawns with \`current_dir(corpus)\`, so the relative path CI passes resolved against the wrong directory. Now canonicalized to absolute at locate time.
- The disparate numbers you flagged were partly a real methodology gap: candidate and baseline ran to completion one side at a time, so runner drift (thermal, noisy neighbor) landed on whichever side ran second instead of hitting both symmetrically. Reworked to interleave rep-by-rep (prepare once per side, then alternate one measured rep at a time), and bumped reps 3→7. `heph-bench run` gained `--skip-prepare`/`--rep-offset`/`--append` to make that possible across separate process invocations.

Rep count was already in both the JSON (`baseline_reps`/`candidate_reps`) and the markdown (`(n=3)` in the ms cell) — confirmed it's there, didn't need a change.

All verified locally end-to-end (interleaved accumulation, merge+compare producing one table, the path fix) before pushing — `cargo test`/`clippy --all-targets` clean, `actionlint` clean.

PR #275, opened 2026-07-30
fix(plugin-go): fail closed when Go SDK checksum is unconfigured

raphaelvigee:
## Summary
- The Go SDK toolchain download (`gotool = "<version>"`) only verified its tarball SHA-256 when the provider's `checksums` option had an entry for that version/platform. Absent one — the default, out-of-the-box case — it downloaded, extracted, and cached the SDK fully unverified beyond TLS, only logging a `tracing::warn!`. This contradicted the module doc's blanket claim of SHA-256 verification and left a real supply-chain gap (a compromised/MITM'd download silently trusted).
- Fixes it by failing closed in `ProviderInner::handle_get` (the toolchain-target resolution path) rather than inside the driver's `run()`: a hermeticity review caught that gating in `run()` alone would let an engine cache hit — which skips `run()` entirely — silently satisfy a since-tightened policy with a stale unverified artifact.
- Adds an explicit `allow_unverified_sdk` provider option (off by default) as the escape hatch for anyone who's accepted the risk knowingly, per product-vision review — matching the bar the sibling `heph-govet` binary provisioning already sets (verified by default, unverified only as a named opt-out).

## Test plan
- [x] `test_get_toolchain_target_fails_closed_with_no_checksum_configured` — proves the default (no `checksums` entry, no opt-out) now errors instead of silently resolving; this is red against the pre-fix code.
- [x] `test_get_toolchain_target_uses_configured_checksum` — a configured checksum still resolves and threads through to the spec.
- [x] `test_get_toolchain_target_allows_unverified_when_opted_in` — `allow_unverified_sdk = true` restores the old warn-and-download behavior.
- [x] `cargo test -p plugin-go --lib` — 404 passed, 0 failed.
- [x] `lint` (clippy + fmt) — clean.

https://claude.ai/code/session_01VGk4reLZL2Ump4LHYwsBro

### DIFF
--- a/crates/plugin-go/src/plugingo/provider.rs
+++ b/crates/plugin-go/src/plugingo/provider.rs
@@ -53,12 +53,6 @@ pub struct Config {
     /// absent here downloads unverified (the toolchain driver warns); supply an
     /// entry to enforce verification. Empty for `gotool = "host"`.
     pub sdk_checksums: HashMap<String, String>,
-    /// Escape hatch: accept a hermetic SDK download with no matching
-    /// `sdk_checksums` entry instead of failing closed. Off by default — the
-    /// toolchain target's [`GetError`] refuses to resolve an unverified
-    /// download unless this is set. Set via the optional `allow_unverified_sdk`
-    /// provider option.
-    pub allow_unverified_sdk: bool,
     /// Addr of the `heph-govet` binary the lint/format targets exec. Defaults to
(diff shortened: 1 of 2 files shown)


raphaelvigee:
Reverting — checksum verification being opt-in is a deliberate prior decision (e995ffda), not a bug. Closing.

PR #284, opened 2026-07-30
fix(bench): canonicalize --corpus in dist::run; fall back when baseline heph-bench lacks --skip-prepare

raphaelvigee:
## Summary

Follow-up to #274. That PR's last two pushes (interleaved measurement + duplicate-header fix, then this fix) landed *after* #274 was already merged — GitHub doesn't retroactively pull in commits pushed to a branch post-merge. This PR carries the one commit that never made it in: `e6efddb3` from that branch (cherry-picked clean onto current `master`; the interleave/merge-per-tier fix and the PostHog integration are already on `master` from #274 itself).

Two real bugs from live CI, both the same root cause pattern (relative path + a `current_dir` change re-interpreting it against the wrong base):

- **Tier B crashed outright**: `read plugin manifest bench-corpus/heph-go-plugin.json: No such file or directory`. `write_go_config` writes that path into `.hephconfig`'s `path:` entry using whatever `--corpus` was given (relative — CI passes plain `bench-corpus`), and `build_go_tree` spawns `heph` with `current_dir(corpus)`. Once that cwd change takes effect, `heph` re-resolves the config's relative path against its own cwd — already inside `corpus` — landing one directory too deep. Same fix pattern as `Dist::locate`'s earlier absolute-path fix: canonicalize `corpus` once at the top of `dist::run`.

- **Tier A failed**: `unexpected argument '--skip-prepare'`. Baseline's `heph-bench` is fetched from a past release, built *before* the PR that added that flag — it doesn't understand it yet. Added a capability probe (grep baseline `heph-bench`'s `--help` for `--skip-prepare`) with a non-interleaved fallback for that one run — self-resolving once a release built after this merges becomes the baseline. Tier B is unaffected: it only ever runs candidate's own `heph-bench` as the orchestrator, never baseline's.

## Test plan

- [x] Verified locally: a relative `--corpus` with a fake `heph` script confirms `.hephconfig`'s `path:` entry is now absolute (previously relative and wrong once resolved from `heph`'s own cwd).
- [x] Verified `--help` output actually contains `--skip-prepare` for the capability-detection grep to match.
- [x] `cargo test`/`clippy --all-targets` clean, `actionlint` clean, YAML parses.
- [ ] Not yet validated by a live CI run against a real baseline missing `--skip-prepare` (the fallback path) — this PR's own merge-base still predates the flag entirely, so it's the fallback path that gets exercised first, not the interleaved path. Flagging as an honest gap rather than claimed coverage.

https://claude.ai/code/session_01F9mbNMBQVSrsJojDYnQcDV

### DIFF
--- a/.github/workflows/heph.yml
+++ b/.github/workflows/heph.yml
@@ -763,9 +763,13 @@ jobs:
   #
   # Two tiers per platform, matching `crates/e2e` vs `crates/bin-e2e`'s split
   # (see `.claude/testing.md`):
-  #   - inprocess (Tier A): the engine linked directly, no process spawn, no
+  #   - inprocess (Tier A): the engine linked directly into `heph-bench`, no
   #     plugin cdylib. Isolates engine-internals cost (cache, hashing, DAG
-  #     resolution) from process-startup/dlopen noise.
+  #     resolution) from dlopen noise, which Tier B measures instead. It DOES
+  #     spawn a process per rep now (on both sides, symmetrically — see
+  #     `run inprocess`'s doc comment in `crates/bench/src/main.rs`), traded
+  #     deliberately for never needing baseline's binary to understand new
+  #     orchestration flags again.
   #   - dist (Tier B): the real prebuilt `heph` binary + real `go` plugin
   #     cdylib, dlopened for real — the only way to measure ABI-crossing and
   #     plugin-load cost. Head's `heph-bench` drives both sides here (the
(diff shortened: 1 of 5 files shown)


raphaelvigee:
Pushed a bigger redesign addressing both points raised:

**Reverted \`--skip-prepare\`/\`--rep-offset\`/\`--append\` entirely** rather than patching around it — that flag design coupled orchestration (warmup/rep counts, interleaving, output format) to the same CLI surface baseline's binary has to understand, so every future orchestration change would need its own bootstrap PR. Already hit that twice.

**New shape**, matching what Tier B already did correctly (candidate's own \`heph-bench\` is the stable orchestrator; the real \`heph\` binary it spawns has a tiny, rarely-changing contract):
- \`heph-bench measure-inprocess --corpus --scenario --mode prepare|once [--mutate-seed N]\` — new minimal, stable "subject" primitive. \`once\` prints *only* the elapsed ms. Small enough it shouldn't need to change again for orchestration reasons.
- \`heph-bench run inprocess\`/\`run dist\` are now the *only* orchestrators — one call per scenario, taking both \`--candidate-*\`/\`--baseline-*\`, doing the full interleaved measurement internally (prepare both sides once, then alternate measured reps). Tier A always spawns \`measure-inprocess\` as a subprocess for *both* sides, even candidate (which technically doesn't need to be spawned) — so both pay identical process-spawn overhead, symmetric noise rather than a fixed advantage to whichever side would otherwise skip it.

Also re: interleaving mechanism — no benefit to reaching for hyperfine here. Its statistical rigor assumes it owns the whole repeated-run loop for one command; given multiple commands it exhausts one fully before starting the next, never alternating. Forcing interleaving via `-m 1` per rep would just add a wrapper-process spawn for zero benefit at n=1.

**Per-scenario/tier granularity** (comparison + PostHog) was already there — \`compare::verdicts()\` already returns one entry per scenario, unchanged by this redesign. Confirmed it survived by running the full \`corpus → orchestrate → merge → compare → PostHog-payload\` pipeline locally and checking scenario tags stayed intact (`["cold","full-hit","incremental"]`, distinct events).

\`heph.yml\` simplifies a lot: one \`heph-bench\` call per scenario now (not per side, not a bash rep loop), and the capability-detection step + its fallback are gone — an incompatible baseline just fails that scenario's run gracefully (already-existing missing-file handling in "Compare and report").

Verified locally end-to-end: \`measure-inprocess\` in isolation, both tiers' full orchestrators (Tier B via a fake \`heph\` script, confirming the absolute-path fix survived), and the full downstream pipeline. \`cargo test\`/\`clippy --all-targets\` clean, \`actionlint\` clean.

PR #291, opened 2026-07-31
ci: split release binaries into std/debug flavours

raphaelvigee:
## Summary
- Linux release binaries were ~200MB because `[profile.release]` shipped DWARF line-table debug info (`debug = "line-tables-only"`, `strip = "none"`) unconditionally.
- Split the release profile into two flavours: **std** (default `[profile.release]`, now `debug = false` + `strip = "symbols"` — small) and **debug** (new `[profile.release-debug]`, keeps the old unstripped behavior for in-process backtrace symbolication).
- CI `build` job now builds both flavours of the `heph` binary per platform and publishes them as separate release assets: `heph_<os>_<arch>` (std) and `heph_debug_<os>_<arch>` (debug). `heph-bench` and the plugin cdylibs are unaffected — one copy of each, built under `std`.
- `.hephconfig` gains an optional `versionFlavour` field (default `""`), mirroring `version`. When set (e.g. `debug`), self-upgrade downloads `heph_<flavour>_<os>_<arch>` instead of the default `heph_<os>_<arch>`.

## Test plan
- [x] `cargo test -p config -p selfupdate` — new tests for `versionFlavour` parsing/merge, and flavoured `binary_name`/`download_url`.
- [x] `cargo clippy -p config -p selfupdate --all-targets -- -D warnings` and `cargo fmt --check`.
- [x] `cargo check --profile release-debug -p config` / `cargo check --release -p config` — confirms both Cargo profiles resolve.
- [x] Workflow YAML validated with `ruby -ryaml`.
- [ ] CI run on this PR (build job timeout bumped 30→45min for the extra build).

https://claude.ai/code/session_013G58Qo3Uyo82F9GRiYrufK

### DIFF
--- a/.github/workflows/heph.yml
+++ b/.github/workflows/heph.yml
@@ -366,13 +366,18 @@ jobs:
           ./$BIN_NAME version
           ./$BENCH_BIN_NAME --help
           # Both flavours launch, and each reports the flavour patch-flavour.sh
-          # actually stamped — not just that the binary runs at all.
+          # actually stamped — not just that the binary runs at all. The
+          # baked-in version already carries its own build metadata (a commit
+          # hash), so check for the debug flavour marker specifically, not
+          # "any build metadata at all" (`hcore::version::reported` joins the
+          # flavour onto the existing metadata with a `.`, or a bare `+` if
+          # there wasn't any — either way it ends in `.debug` or `+debug`).
           std_version="$(./$BIN_NAME version)"
           debug_version="$(./$DEBUG_BIN_NAME version)"
           echo "std: $std_version"
           echo "debug: $debug_version"
-          case "$std_version" in *+*) echo "std flavour must not report build metadata: $std_version" >&2; exit 1 ;; esac
-          case "$debug_version" in *+debug) ;; *) echo "debug flavour did not report +debug: $debug_version" >&2; exit 1 ;; esac
+          case "$std_version" in *+debug|*.debug) echo "std flavour must not report the debug flavour: $std_version" >&2; exit 1 ;; esac
+          case "$debug_version" in *+debug|*.debug) ;; *) echo "debug flavour did not report itself: $debug_version" >&2; exit 1 ;; esac
         env:
           BIN_NAME: "heph_${{ matrix.os }}_${{ matrix.arch }}"
           BENCH_BIN_NAME: "heph-bench_${{ matrix.os }}_${{ matrix.arch }}"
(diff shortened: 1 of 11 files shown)


raphaelvigee:
Pushed a follow-up commit addressing two review points:

- **CI now compiles `heph` once**, not twice — the "std" flavour is derived from the same build via `zig objcopy --strip-all` (already in the toolchain for zigbuild, LLVM-based so it handles the cross-compiled Linux arm64 output from an amd64 runner, unlike host `strip`/binutils). Roughly halves the added build cost from the original approach.
- **Fixed a gap**: a workspace with `version: X` that adds/changes `versionFlavour` (same `X`) now actually triggers a re-fetch. The running binary has no compile-time way to know its own flavour (single compile now backs both), so `decide()`'s version-only comparison would've silently treated a flavour-only edit as a no-op. Fixed by tracking each workspace's last-confirmed `(tag, flavour)` in `~/.heph/pins/<hash>`, written only after the re-exec'd child proves `execve` succeeded — never by the parent before a download/exec it hasn't confirmed, so a failed exec can't leave a false "already applied" record blocking retries.

New tests: `check_pin_*` (the flavour-drift decision logic) and `pin_file_*` (roundtrip) in `crates/selfupdate/src/lib.rs`.

raphaelvigee:
Follow-up: swapped the pin-state-file approach for something simpler — post-build flavour stamping.

`crates/core/src/version.rs` reserves a fixed-width byte slot (`FLAVOUR_SLOT`) baked into the binary at compile time with a distinctive marker. `scripts/patch-flavour.sh` overwrites it in place after the build (`std` → empty, `debug` → `"debug"`), no second compile. `hcore::version::flavour()` reads it back at runtime, and `heph version` now reports it as semver build metadata (`v1.2.3+debug`). Self-upgrade's `decide_pin` compares the running binary's own flavour directly against the config's `versionFlavour` — no more `~/.heph/pins/` marker file, no exec-confirmation dance, since the running binary can just answer the question itself now.

Two real bugs found and fixed while verifying against an actual `--release` build (not just `cargo build`/dev profile, where neither reproduced):
- An immutable `static` with a fully-known initializer is exactly what LLVM constant-folds away under thin-LTO + `opt-level = 3` — the patched bytes were silently ignored until the read went through `std::ptr::read_volatile`, which the language guarantees can't be folded.
- `patch-flavour.sh` used `${2:?...}`, which bash treats as "unset" for an empty string — rejecting the std flavour's legitimately-empty flavour argument.

Also added a `bin-e2e` assertion (`shipped_binary_launches_outside_a_workspace`) that the shipped std-flavour binary's `heph version` output carries no build metadata — this is a shipped-artifact-pipeline property (patch + strip), not something an in-process test can see.

PR #292, opened 2026-07-31
feat(plugin-js): js_bundle driver, esm/cjs + node/browser variants

raphaelvigee:
M6 of the JS/TS heph plugin plan (part 6/6, top of the stack) — the final milestone, plus a first performance baseline for the whole plugin.

A cacheable js_bundle ManagedDriver wrapping esbuild, with format (esm/cjs) and target (node/browser) as plain addr args resolved with a flat default rather than porting the Go plugin's ancestry/universe variant machinery — this plugin's addr model is a single flat workspace, so there's no ancestor chain or cross-subtree variant-pin problem for that machinery to solve (researched and confirmed against `crates/plugin-go/src/plugingo/variant.rs` and this crate's own flat `workspace.rs` model before deciding).

Whole-graph cache key by design (the entry point's full transitive closure via ImportGraph::runtime_edges, cross-package recursion unlike js_test's one-hop trim), third-party deps resolved via the same lockfile-driven mechanism as every other driver, tsconfig (plus its extends chain) declared and staged the same way js_typecheck does.

Reviewed by feature-quality/code-quality/hermeticity. Two functional BLOCKERs found and fixed: the discovered third-party import closure was never wired into esbuild's `--external` flags, so bundling any package with a real npm runtime dependency failed outright; and the output directory didn't vary by format/target, so esm and cjs variants of the same package collided on the same declared output path. A third BLOCKER (missing tsconfig Input) meant path-aliased/JSX/decorator-configured TS entry points silently failed or cached wrong. All three fixed with regression tests, including a real-esbuild end-to-end proof for the external-deps fix.

Explicitly deferred (disclosed in module docs): `target=browser` doesn't yet change what's resolved (only what esbuild is told to assume) since the resolver has no browser condition/main-field support; a config-reference-scanning helper shared with js_test/js_lint reads a config-referenced file before checking workspace containment rather than after (dormant for js_bundle's only supported config format today, live for js_test already — flagged as a cross-cutting follow-up); rollup/webpack/vite bundlers.

**Open item for a future PR**: the design doc's v1 scope table lists six drivers including `js_format`, but the Milestones list only ever defined M0-M6 for the other five (install/typecheck/test/lint/bundle) — `js_format` has no milestone and isn't built yet.

## Performance baseline

Speed was the plugin's stated #1 design goal from the start — this had never actually been measured until now. Added `#[ignore]`d ad-hoc benches (never run under `tst`/default `cargo test`) measuring `Provider::import_graph`'s cold cost (first parse+resolve per package) vs warm cost (the M5 per-package memoization) on synthetic 20- and 150-package workspaces, plus a parse-vs-resolve phase split.

Results (darwin/arm64, this machine only — not transferable to other targets without re-running there):
- **Cold**: ~2.5-3ms/package, ~0.16-0.2ms/file (150 packages × 15 files: median 385ms, range 352-679ms across 8 runs — this machine's noise floor is large).
- **Warm**: ~500ns total across 150 packages — a cache hit never re-enters the build path at all, confirming M5's fix delivers as designed.
- **Phase split**: resolution (`oxc_resolver`) is 97%+ of cold cost; parsing (`oxc_parser`) is noise by comparison (~2.6-2.7%).

No prior baseline existed, so this establishes one rather than comparing against one — no regression, no obvious hotspot found. Real gap flagged: there's no JS/TS scenario in `crates/bench`/`crates/bench-corpus` (Go-only today), so this path has no standing CI perf-regression coverage yet — building that out is the natural follow-up.

With this PR, M0-M6 of `ai-docs/js-plugin-plan.md` are all implemented across the stack below.

## Test plan
- [x] `cargo build -p plugin-js -p plugin-js-cdylib`
- [x] `cargo test -p plugin-js` (319 passed, 18 ignored — require real `tsc`/`vitest`/`jest`/`oxlint`/`esbuild` binaries not provisioned in this devenv)
- [x] `cargo clippy -p plugin-js -p plugin-js-cdylib --all-targets -- -D warnings` (clean)
- [x] `cargo fmt --check -p plugin-js -p plugin-js-cdylib`
- [ ] CI (`tst`, `lint`) on this PR

https://claude.ai/code/session_01M3wZfyPsG8stfRQuybLRjN

### DIFF
--- a/crates/plugin-js-cdylib/src/lib.rs
+++ b/crates/plugin-js-cdylib/src/lib.rs
@@ -1,6 +1,6 @@
 //! The JS/TS plugin as a loadable cdylib behind the stable ABI.
 //!
-//! **M0-M5 scope**: exports the `js` provider (package discovery + pnpm/npm
+//! **M0-M6 scope**: exports the `js` provider (package discovery + pnpm/npm
 //! workspace-member resolution + lockfile-driven dependency wiring + oxc-based
 //! import-graph resolution), its `js_package_info` dependency-wiring driver,
 //! the hermetic `js_install` third-party fetch driver, the `js_typecheck`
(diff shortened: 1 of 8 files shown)


raphaelvigee:

PR #302, opened 2026-07-31
feat(plugin-sdk,plugin-stabby): spawn plugin and host bodies at the ABI seam, symmetrically

raphaelvigee:
Stacked on #300 (`raphaelvigee/provider-list-guard`) — the executor-carried `IN_PROVIDER_LIST` guard is the prerequisite that makes spawning provider bodies safe.

## Design

A cdylib statically links its own tokio; a plugin body polled by a host worker (or a host callback body polled by a plugin worker) that touches its reactor/timer panics ("there is no reactor running"), and a panic across the `extern "C"` seam is a non-unwinding process **abort**. Until now only driver `run` and hook drains were spawned onto the right runtime — every other entry-point body ran inline on the peer's workers.

This PR makes the spawn **symmetric at the seam**:

- **Guest** (`plugin-sdk::serve`): every ABI entry-point body — provider `get`/`probe`/`list`/`list_packages`/`call_function`, driver `parse`/`apply_transitive` — spawns onto the plugin's own runtime (workers named `heph-plugin-worker`); the returned `DynFuture` only awaits the `JoinHandle` (runtime-free to poll; `hook_on_events` was the in-tree precedent).
- **Host** (`plugin-stabby::host`): `HostExecutor::{result,query,states_under}` and `HostFunctionRegistry::call_registered` capture `Handle::current()` + `Span::current()` at `wrap()` time and spawn their bodies there, `.instrument()`ed. `note_dep` stays synchronous (ready future, reactor-free). Without this mirror, the first uncached dep resolved through a real cdylib aborts the process — in-process tests structurally cannot catch it (host and plugin share one image); only the bin-e2e dlopen seam can.
- **Mechanics**: `JoinHandle` (tokio's task harness is the `catch_unwind`), never `resume_unwind` in the wrapper (its poll runs inside the extern shim); panic → `plugin <name>: <method>(<key>) panicked: <payload>` error body; abort-on-drop guard around the handle (drop is the only stop signal for the entry points with no `CancelRegistry` wiring; the cooperative `await_with_cancel` path is untouched); new `hcore::hmemoizer::spawn_on_with_cycle_ctx(&Handle, fut)` used at every seam spawn on both sides (and retrofitted onto `run_bidi`'s existing spawn) so cycle-detection frames survive the spawn; the guest wrapper arms `hcore::blocking::Backstop` on every pending poll (hazard not reproduced in isolation — defense-in-depth consistent with `blocking::run`, see docs/CONCURRENCY_MEASUREMENTS.md §2).

## Review-board verdicts

Design (consolidated v2):

- product-vision: **SHIP WITH CHANGES** (panic error shape, named threads, diagnosability) — incorporated.
- compatibility: **BREAKING-as-proposed** — fixed by the prerequisite PR #300 (task-local list guard moved onto the executor instance).
- feature-quality: **BLOCKED-as-proposed** — fixed by the symmetric host mirror + the Tier B cost gate below.
- code-quality: **PASS WITH NITS** on this symmetric shape.

Implementation review (second pass, addressed in the review-fixes commit):

- compatibility: **COMPATIBLE**, no ABI bump.
- code-quality: **PASS WITH NITS** — majors fixed: multi-token `inflight` (`HashMap<String, Vec<Token>>`; a request cancel trips every concurrent call under the shared `req-{n}` id, not just the last registrant), explicit `wrap(handle)`/`wrap_inline` constructors instead of a silent `try_current` probe, purpose-made `provider_functions` span for the once-per-process registry wrapper, `panic_text` deduped into `hplugin_stabby::seam`, `<unnamed>` fallback for a failed `config()`, bounded `precancelled` (consumed by a cancelled guard's drop), doc-citation reword, `cancel_abandoned` comment addendum for the seam's async-abort exception.
- feature-quality: **BLOCKED** — both blockers fixed: a bin-e2e test crossing the host mirror over the real dlopen'd go cdylib (`states_under` callback from a plugin worker; sibling-declared variant makes the assertion non-vacuous), and deterministic `CancelRegistry` unit tests (each verified red against the pre-fix registry). `HostTask` now arms the Backstop on pending polls (its completion wake crosses the seam toward a guest worker — the guest-side backstop cannot cover it) and has its own abort-on-drop test.

## Semantic deltas

- **Eager start**: `invoke()` now runs guest code before the wrapper's first poll (request decode, `CancelRegistry::enter`, `note_dep` inserts). Documented as a seam invariant on the dispatch impls. `CancelRegistry::enter` stays inside the spawned body so pre-cancel parking still applies. DepDag note: an abandoned-pre-poll body can now have inserted edges before its wrapper is dropped; edges live in the per-request DepDag, which dies with the request, and re-issued calls re-insert idempotently.
- **Drop-cancel is now asynchronous**: the abort lands on a plugin worker after the host's drop returns, where the old inline body died synchronously with the destructor.

## Latent races fixed (found by the new tests / flagged by review)

- `CancelRegistry` `enter`/`cancel` each *checked* the other's map before *publishing* to its own — a cancel landing between a call's precancelled check and its inflight insert was lost. Hittable once spawn is eager; the race-loop test caught it at iteration 2. Both sides now publish before checking.
- `CancelGuard::drop` removed its request id unconditionally, but request ids name an engine request, not a call — a stale guard's late drop could deregister a live successor's token. Removal is now identity-checked (`StdCancellationToken::ptr_eq`).

## Cost

~2 spawns per target per warm run (get/parse ride per-run memoizers). **Tier B bench pending (gate)**: real-cdylib cold (go corpus) + full-hit warm, baseline vs candidate, run serially on this PR before merge; results go into docs/CONCURRENCY_MEASUREMENTS.md. If the warm delta is real, the spawn scopes down to reactor-needing entry points only. Guest runtime sizing (currently ncores, justified by driver `run`) is a visible decision surfaced for perf-measurement sign-off with those numbers.

## Tests

- `serve::tests::panicking_provider_body_is_contained_as_error` — panic → actionable error (plugin, method, key, payload), process survives.
- `serve::tests::dropped_seam_future_aborts_the_spawned_body` — guest abort-on-drop: side-effect flag + deadline poll.
- `host::tests::dropped_callback_future_aborts_the_spawned_host_body` — host-mirror abort-on-drop, same shape.
- `serve::tests::precancel_racing_the_spawn_is_never_lost` — 100-iteration race loop; caught the enter/cancel TOCTOU.
- `serve::tests::registry_cancel_before_enter_starts_cancelled`, `registry_cancel_after_enter_trips_the_token`, `registry_stale_guard_drop_does_not_deregister_successor`, `registry_cancel_trips_every_concurrent_call_under_one_id`, `registry_cancelled_guard_drop_leaves_no_state_behind` — deterministic registry contract; the stale-drop and multi-token tests verified red against the pre-fix registry.
- `serve::tests::nested_seam_get_completes_without_deadlock` — get → executor.result(dep) → second get spawned into the same plugin runtime; completes under timeout. Exercises the host mirror's spawn path too.
- `guest::tests::host_callback_panic_surfaces_as_error_not_abort` — host-side mirror of panic containment.
- bin-e2e `plugin_dylib::shipped_go_cdylib_list_calls_back_states_under_across_the_seam` — the host mirror over the REAL dlopen'd go cdylib: `list` calls back `states_under` from a plugin worker; the `release` variant declared only at the sibling `//cmd` makes `//lib:build_lib@v=release` provable only via the callback.
- Existing suites green: plugin-sdk, plugin-stabby, core, engine, e2e; plugingo-e2e build suite (16/16). `cancellation_propagates_to_plugin` green unchanged.

## Follow-ups (recorded, not in this PR)

- A panic inside a lazily-pulled list iterator still unwinds through the extern `GuestItemStream::next` → abort (pre-existing). Fix shape: `catch_unwind` in `frame_iter`, mapping to a `stream_err` frame.
- `run_bidi`'s fire-and-forget run task surfaces a panic only as a channel close ("run stream ended without a result"). Fix shape: await its `JoinHandle` and emit the panic text as the terminal error frame.
- A bin-e2e deliberate-panic fixture cdylib, so panic containment (not just reactor safety) is also proven across the real dlopen seam.

https://claude.ai/code/session_014wDi9n2Fi1aN6jwy8ET7WG

### DIFF
--- a/crates/bin-e2e/tests/plugin_dylib.rs
+++ b/crates/bin-e2e/tests/plugin_dylib.rs
@@ -80,6 +80,65 @@ fn shipped_gha_cdylib_loads() {
     );
 }
 
+/// The spawn-at-the-seam host mirror, across a REAL cdylib: the go provider's
+/// `list` calls back into the engine executor (`states_under`, the module
+/// variant universe) from a plugin-runtime worker. In-process tests
+/// structurally cannot cover this — host and plugin share one tokio image
+/// there, so the callback body always finds a runtime to run on. Here the
+/// callback future is built by the host binary and polled by the plugin's own
+/// workers: exactly the asymmetry that panics ("no reactor running") and then
+/// aborts at the extern seam if the host side does not spawn the body onto the
+/// engine runtime.
+///
+/// The fixture mirrors plugingo-e2e's `variant_sibling`: the `release` variant
+/// is declared ONLY at `//cmd`, so it is absent from `//lib`'s ancestry and
+/// `//lib:build_lib@v=release` can be listed only if the module universe came
+/// back through the `states_under` callback — the assertion cannot pass
+/// vacuously. `gotool: host` + query-only keeps the fixture offline: listing
+/// runs no toolchain.
+#[test]
+fn shipped_go_cdylib_list_calls_back_states_under_across_the_seam() {
+    let dist = Dist::locate();
+    let ws = Workspace::new().expect("workspace");
+    let dylib = dist.plugin("go");
+    assert!(dylib.is_file(), "missing {}", dylib.display());
+
+    let manifest = ws.root().join("heph-go-plugin.json");
+    let sum = sha256_file(&dylib).expect("hash go cdylib");
+    write_manifest(&manifest, "go", &dylib, Some(&sum)).expect("write manifest");
+
+    ws.config(&format!(
+        "{BASE_CONFIG}  - path: {}\n    options:\n      gotool: \"host\"\n",
+        manifest.display()
+    ))
+    .expect("write config");
+
+    ws.write("go.mod", "module example.com/seam\n\ngo 1.21\n")
+        .expect("write go.mod");
+    ws.write(
+        "lib/lib.go",
+        "package lib\n\nfunc Greet() string { return \"hi\" }\n",
+    )
+    .expect("write lib.go");
+    ws.write("cmd/main.go", "package main\n\nfunc main() {}\n")
+        .expect("write main.go");
+    ws.write(
+        "cmd/BUILD",
+        "provider_state(\n    provider = \"go\",\n    variants = {\"release\": {\"goos\": \"linux\", \"goarch\": \"amd64\"}},\n)\n",
+    )
+    .expect("write BUILD");
+
+    let out = ws.run(&dist, &["query", "-e", "//lib/..."]).expect("run");
+    assert!(out.status.success(), "{}", describe(&out));
+    let stdout = String::from_utf8_lossy(&out.stdout);
+    assert!(
+        stdout.contains("//lib:build_lib@v=release"),
+        "the sibling-declared variant must come back through the plugin's \
+         states_under callback across the seam: {}",
+        describe(&out)
+    );
+}
+
 /// Regression test for the log-sink-before-`create` ordering bug: if the host
 /// installs its log sink *after* calling the plugin's `create`, a
 /// `tracing::error!` logged during construction failure has no subscriber to
--- a/crates/core/src/hasync/cancellable_std.rs
+++ b/crates/core/src/hasync/cancellable_std.rs
@@ -69,6 +69,13 @@ impl StdCancellationToken {
     pub fn is_cancelled(&self) -> bool {
         self.inner.cancelled.load(Ordering::SeqCst)
     }
+
+    /// Token identity: `true` iff `other` is a clone of `self` (they share the
+    /// same underlying state). Lets a registry keyed by a reusable id confirm an
+    /// entry is *this* token before acting on it.
+    pub fn ptr_eq(&self, other: &Self) -> bool {
+        Arc::ptr_eq(&self.inner, &other.inner)
+    }
 }
 
 /// Future returned by [`StdCancellationToken::cancelled`]. Resolves once the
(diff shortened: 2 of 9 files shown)


raphaelvigee:
## Tier B bench (merge gate) — PASS

Serial local runs on macOS arm64 (10-core M-series), machine otherwise idle. Baseline dist = `debe3f65` (PR base head), candidate = `a362b926` (this PR head); both built `--release --locked` from clean target dirs, staged + `macos-portable.sh`'d exactly like `e2e`. Corpus = CI perfbench params (seed 1, 4000 bash targets / 200 packages, 60 go packages, layers 10, fan-out 4), `gotool: host`. Candidate `heph-bench run dist` orchestrated both sides, reps interleaved.

| Run | Scenario | Baseline (ms) | Candidate (ms) | Delta | Threshold | Verdict |
|---|---|---:|---:|---:|---:|---|
| 1 | cold (n=5) | 198.1 | 288.4 | +45.6% | 26.3% | REGRESSION |
| 2 | cold (n=7) | 202.0 | 199.6 | −1.2% | 29.1% | ok |
| 3 | cold (n=7) | 198.8 | 189.7 | −4.6% | 36.3% | ok |
| 1 | full-hit (n=7) | 135.4 | 137.1 | +1.2% | 10.1% | ok |

Run 1 started the moment two back-to-back release builds finished; its +45.6% did not reproduce in either settled run. That is the interleaving/thermal trap the load-harness docs warn about, not a signal. Verdict: the symmetric spawn at the ABI seam costs nothing measurable on the real dlopen'd cold path, and the warm (full-hit) path is unchanged.

PR #304, opened 2026-08-01
feat(engine): flip request memoizers to the task-backed cell

raphaelvigee:
The measured flip PR of the memoizer redesign: every RequestState
memoizer (spec/def/meta/result/locked_result/execute_cache/...) now
spawns its cold computations as tasks on the engine's runtime, captured
once at Engine::new (which now requires a runtime context — it fails
loudly at construction rather than at first request). The poll-cell
machinery stays in-tree and unreferenced by the engine; deleting it is
the follow-up PR, so reverting this one is this commit alone.

What changes at runtime: a warm deep descent is task hops instead of one
synchronous poll per chain (~100KiB/level — the 2MiB-stack overflow
GrowStack exists for), depth no longer serializes on one worker, and
cancellation tears down via abort per cell instead of a recursive drop
cascade. Teardown becomes asynchronous: permits and interests drain when
the runtime processes the aborts — the task-mode wedge test asserts the
permit comes back bounded-eventually and a live acquire is served while
the chain drains.

deep_warm_chain_completes_on_a_2mib_stack pins the stack claim: 200-deep
warm chain on an explicit 2MiB thread. Verified red against the poll
cell with GrowStack neutered (SIGABRT, stack overflow) and green under
the flip — it is the regression gate for deleting GrowStack later.

Providers (plugin-go pkg caches, buildfile pkg_cache) deliberately stay
on the poll cell: flipping them is a separate, separately-measured PR so
this one's bench verdict attributes to the engine alone.

Full `tst` suite green under the flip.

---


raphaelvigee:
## Bench gate — numbers in; two of the design's own stop-gates trip; decision flagged, not taken

Serial interleaved runs, macOS arm64 (10-core), machine otherwise idle. Baseline = #303 head `0ffa0090` (poll cells), candidate = this PR's `49dae4c2` (task cells). Both sides built `--release --locked` from clean target dirs; candidate `heph-bench` orchestrated every comparison rep-by-rep. Corpora: **std** = CI perfbench shape (4000 targets / 200 pkgs / 10 layers / fan-out 4 / 60 go pkgs), **deep** = 4000 / 200 / **27 layers** / fan-out 2, **thin** = 200 targets / **200 layers** / fan-out 1.

| Tier | Corpus | Scenario | Baseline (ms) | Candidate (ms) | Delta | Harness verdict |
|---|---|---|---:|---:|---:|---|
| A | std | full-hit (n=8) | 541.0 | 588.3 | +8.7% | ok |
| A | std | **full-hit (n=20)** | 538.8 | 581.3 | **+7.9%** | ok |
| A | std | cold (n=5) | 78754 | 65859 | **−16.4%** | ok |
| A | std | incremental (n=6) | 515.2 | 562.7 | +9.2% | ok |
| A | deep | full-hit (n=8) | 486.9 | 525.7 | +8.0% | ok |
| A | deep | **cold (n=5)** | 309394 | 57787 | **−81.3%** | ok |
| A | thin | **full-hit (n=8)** | 106.3 | 47.1 | **−55.7%** | ok |
| B | std | cold (n=5) | 201.2 | 209.6 | +4.2% | ok |
| B | std | full-hit (n=7) | 138.9 | 141.7 | +2.0% | ok |

Max RSS, deep-corpus warm rep (`/usr/bin/time -l`, repeated twice per side, stable): baseline ~200MB, candidate ~302MB (**+50%**). Attribution runs: std +72MB (+35%), thin **−16MB (−25%)** — the delta scales with target count, not depth, and reverses on small graphs. It is peak-concurrency working set, not a leak: the task model holds more cells live simultaneously — the same parallelism that produces the −81% cold and −56% deep-warm wins.

### The honest verdict

The harness passes everything (its noise-calibrated thresholds are wide). But the design doc set stricter stop-gates, and two trip:

- **full-hit +7.9% at n=20** (gate: ~2–3%). Stable across n=8/n=20 and three corpora; ~+10µs/target — almost exactly the predicted 4–6 request-cold spawns/target. This is the real price of task-per-cell on the sacred path: +42ms on a 4000-target no-op build.
- **RSS +35–50% at 4000 targets** (gate: ~10%). ~+18–25KB/target peak; reverses to a win on small/deep-thin graphs.

Bought with them: **cold builds on deep graphs collapse from 5.2 min to 58 s (−81%)**, deep warm descents −56%, std cold −16%, and the structural wins (2MiB-stack safety without `stacker`, abort-based cancellation, the deletion PR this enables).

Per the review-board rules this is not mine to wave through: the low-overhead promise is a design pillar and the gates were set deliberately. Options, roughly in order of appeal:

1. **Bound concurrent cold-cell spawns** (semaphore in the task cell's spawn path): caps peak RSS and likely trims the full-hit spawn cost, at some cost to the cold-build parallelism win. Measurable with the exact harness runs above.
2. **Accept the trade** as measured (fast machines, cold builds are where users wait) and record it in the commit body.
3. **Reject the flip** — keep the task cell unused (PR #303 stands alone), revisit later.

Holding here; the stack stays open. All raw jsons + corpora preserved in the bench scratch dir for re-runs.

raphaelvigee:
Correction on option 1 before anyone reaches for it: a semaphore on cold-cell **spawns is the diamond-deadlock class this engine already escaped once** — a parent's body awaits its child's cell, so capping concurrently-live cells deadlocks any chain deeper than the cap (same shape as worker-permit-before-deps, `execute.rs`). The safe levers are narrower than "bound the spawns":

- shrink the per-cell footprint (the `BodyTask` capture set, Notify/JoinHandle slots) — grinds at the ~18–25KB/target, cannot remove the concurrency-inherent part;
- bound only *leaf* work — already what `WorkerPool` does for executes; the metadata cells' RSS is the cost of holding the graph concurrently at all.

So the real menu is: accept the peak-RSS/full-hit price as the cost of the −81%/−56% wins (record it in the commit body), fund a footprint pass first, or reject the flip. Not silently semaphore it.

PR #305, opened 2026-08-01
feat(plugins): flip provider memoizers to the task-backed cell

raphaelvigee:
Completes the memoizer flip for the provider side: the buildfile
provider's pkg_cache and the go provider's four caches (pkg, pkg_addrs,
libs, import_closure) now spawn cold computations as tasks, on a runtime
that is handed to them — never discovered:

- PluginInit gains `runtime`: the engine's handle, injected into every
  in-process provider constructor (buildfile via from_options/new, go via
  from_options/with_config/new). Factory closures pass init.runtime.
- The go cdylib constructs its provider with the cdylib runtime's own
  handle, via the new `plugin_sdk::stabby::cdylib_runtime_handle()` —
  heph_plugin_create runs on a host thread where Handle::current() would
  be wrong or absent, so the export exists precisely to avoid discovery.
- Sync tests that construct providers get a shared static test runtime
  (same shape as the engine's test_rt_enter from the flip PR).

With this, every memoizer in the tree constructs task-backed except the
hcore/engine test fixtures that exercise the poll cell directly — those
die with the poll cell in the deletion PR.

---

### DIFF
--- a/crates/bench/src/inprocess.rs
+++ b/crates/bench/src/inprocess.rs
@@ -59,6 +59,7 @@ fn build_engine(root: &Path) -> Result<Arc<Engine>> {
     e.register_provider(|init| {
         Box::new(heph::pluginbuildfile::Provider::new(
             init.root.to_path_buf(),
+            init.runtime.clone(),
         ))
     })
     .context("register pluginbuildfile provider")?;
--- a/crates/e2e/tests/cache_load.rs
+++ b/crates/e2e/tests/cache_load.rs
@@ -115,8 +115,13 @@ fn build_engine(root: &Path, remote_uri: &str, parallelism: Option<usize>) -> Ar
         ..Default::default()
     })
     .expect("engine");
-    e.register_provider(|init| Box::new(pluginbuildfile::Provider::new(init.root.to_path_buf())))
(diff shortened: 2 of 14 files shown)


raphaelvigee:
## Bench — clean pass

Serial interleaved runs, same rig/method as #304's session. Baseline = #304 head (engine flip, providers on poll cells), candidate = this PR (providers task-backed). Corpus = the CI-shape std corpus (4000 bash targets, 60 go packages).

| Tier | Scenario | Baseline (ms) | Candidate (ms) | Delta | Verdict |
|---|---|---:|---:|---:|---|
| B (real dlopen'd go cdylib) | cold (n=5) | 613.3 | 196.5 | **−68.0%** | ok |
| B | full-hit (n=7) | 151.9 | 154.9 | +2.0% | ok |
| A (in-process; covers buildfile pkg_cache) | full-hit (n=10) | 663.1 | 650.1 | −2.0% | ok |

The go provider's per-package memoized walks stop serializing on their driver poll — that's the −68% cold. Warm paths flat: the provider caches hold a handful of cells per run, so the per-cold-spawn cost that showed on #304's request memoizers (thousands of cells) doesn't register here. No stop-gate trips.

PR #307, opened 2026-08-01
perf(hmemoizer): shrink the task cell's footprint

raphaelvigee:
Two mechanical reductions at hundreds-of-thousands-of-cells scale:

- done + poison merge into one `outcome: OnceLock<Result<V, &str>>` — a
  cell has exactly one terminal state; `is_done` stays value-only
  (poison is deliberately not "done", so a poisoned cell can still be
  evicted by its unwinding joiner and recomputed).
- The cells and graves maps move behind ONE `Arc<Maps>`: every spawned
  body clones a single pointer instead of two, the inventory holds a
  single `Weak`, and the memoizer allocates one shared block instead of
  two.

Also adopts the cross-thread canceller regression test suggested in the
grave-protocol review (repro_evict_to_grave_window_overlaps_bodies): a
canceller parked between eviction and grave-park — the exact window the
one-critical-section fix closed — driven from its own OS thread while a
successor races in.

---

### DIFF
--- a/crates/core/src/hmemoizer/task_cell.rs
+++ b/crates/core/src/hmemoizer/task_cell.rs
@@ -62,11 +62,13 @@ use rustc_hash::FxHashMap;
 /// once `done` is set (at which point `task` is empty — a completed cell pins
 /// no dead task harness).
 pub(crate) struct TaskCell<V> {
-    done: OnceLock<V>,
-    /// Terminal failure that is not a value: the task was dropped without
-    /// publishing (a panicking body outside `once()`'s guard, or its runtime
-    /// shut down). Waiters surface it loudly instead of parking forever.
-    poison: OnceLock<&'static str>,
+    /// The terminal state: `Ok` is the published value; `Err` is poison — the
+    /// task was dropped without publishing (a panicking body outside
+    /// `once()`'s guard, or its runtime shut down), and waiters surface it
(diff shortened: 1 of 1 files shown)


raphaelvigee:
## Measurement — honest null on RSS; keep for hygiene + the adopted repro test

Same rig/method as the stack's earlier sessions. Baseline = #306 head, candidate = this PR. Deep corpus (4000 targets / 27 layers), warm rep, `/usr/bin/time -l`, two runs per side:

| Side | Max RSS |
|---|---|
| baseline | 291 / 300 MB |
| candidate | 296 / 304 MB |

Tier A full-hit (n=10, interleaved): +1.7% — noise.

**The cell-field trims (~40B/cell: merged terminal-state OnceLock, single shared-maps Arc) are invisible against the ~300MB peak**, which is what the earlier attribution already said it was: tokio task harness + the engine's future state machines × cells live at once — the parallelism's working set. The next real levers are engine-side (shrinking `inner_result_addr`-family state machines held across awaits) or accepting the peak as priced in; per-cell shaving is exhausted.

What this PR still carries on its merits: one terminal-state slot instead of two (a cell has one terminal state — that's correctness-adjacent hygiene), one `Arc` clone per spawned body instead of two, and the cross-thread-canceller regression test the grave-protocol review asked for (`repro_evict_to_grave_window_overlaps_bodies`) — pinning the evict/park critical-section fix from its own thread.

PR #313, opened 2026-08-01
refactor(core): run blocking work on tokio's pool, bounded, and delete the backstop

raphaelvigee:
Replaces `hcore::blocking`'s hand-rolled OS-thread pool with `tokio::task::spawn_blocking`, and deletes the machinery that existed only because tokio's own path couldn't be used or trusted.

## Why now — both premises are gone

`hcore::blocking` was `2 * cores` named OS threads behind a crossbeam queue, plus a process-wide `PENDING` waker registry re-woken every 250ms by a `heph-blocking-wake` ticker thread. Two reasons justified that; neither survives:

- **No runtime context.** A cdylib plugin's futures used to be polled by host workers, where the plugin's statically-linked tokio saw no runtime at all — `spawn_blocking` there panics, and a panic crossing the `extern "C"` seam is a non-unwinding process abort. Spawn-at-the-seam (#302–#308) ended that: ABI entry-point bodies run as tasks on the cdylib's own runtime, host callback bodies as tasks on the host runtime. Callers of `run` now have a runtime context on both sides.
- **Distrusted wake-ups.** The backstop insured the result wait against tokio's cross-thread waker "dropping wake-ups on macOS under load". `docs/CONCURRENCY_MEASUREMENTS.md` (checked in) could not reproduce that across ~40M wakes on the pinned tokio, and re-measured the `block_in_place` 0.94→0.74 concurrency regression cited alongside it at parity within noise. The task-backed memoizer already trusts tokio's waker path for every build.

**Net −163 lines**, and the deletions are the point: pool threads, queue, `PENDING`, ticker, `Backstop`, `flush_backstop`, and the GC waker-release coupling.

## The GC coupling dissolves

`flush_backstop` existed because a registered `Waker` could own an `Arc<hmemoizer::Cell>` whose memoized `mem_locked_result` value *is* an addr's riding cache read — so a registration outliving its wait pinned the very guard the post-write trim needs, and the trim batch had to force the registry to let go (three flushes, one per phase). This coupling was flagged in the architecture audit as the most fragile thing in the codebase.

A `JoinHandle` await parks its waker in the task's join slot for exactly the wait's lifetime, and an abandoned request's chains are torn down by the task-backed memoizer's abort cascade, which drops the guards with them. Nothing retains a guard its owning task hasn't kept alive, so there is nothing to flush. The batch keeps its probe / re-probe / one delayed retry — the release edge is now the cascade, still asynchronous to the cleaner thread — and `TrimBatchReport::flushed` goes with the mechanism it measured.

## Sharp-edge decisions

| Edge | Decision |
|---|---|
| **No-runtime contract** | **Changed.** Every `blocking::run` call site was audited; all are post-seam paths polled with a runtime **except the buildfile LSP**, whose `starlark_lsp` request threads are plain `std::thread`s. Fixed by entering the handle the LSP context already captures. The two no-runtime tests are deleted (they pinned a retired contract) and replaced by a test driving the LSP listing from a thread with no ambient runtime. |
| **Concurrency bound** | **Preserved, not widened.** Tokio's blocking cap is `8 * cores + 64`; a `Semaphore` sized exactly like the old pool (`(cores * 2).max(4)`) sits in front. Permit acquired in async land (a waiter dropped while queueing leaves cleanly, having spawned nothing), then rides *into* the job so a detached job still counts against the bound. Right-sizing is a later, measured change. |
| **`PKG_EVAL_SLOTS`** | Invariant kept, restated against the semaphore: a `LoadRegistry` claim waiter parks its run slot while its holder needs one, so evaluations stay strictly below `concurrency_limit()`. Thread supply is no longer the binding resource; run slots are. |
| **Panics / cancellation** | Unchanged semantics. Panics stay transparent (tokio's harness is the `catch_unwind`, payload `resume_unwind`-ed on the caller). Dropping the future still *detaches* a running job — `spawn_blocking` can't be cancelled mid-run, and `JoinHandle` drop gives the same run-to-completion callers rely on to release permits moved into a job. |

## Test evidence

The old pool's threads were the only ones named `heph-blocking-*`, which five "runs off the runtime workers" tests asserted on. Tokio names all its threads alike, so they now record a thread-local `in_blocking_job()` — set only inside a `run` job, so a bare `spawn_blocking` or `tokio::fs` op reads false and can't satisfy a witness by dodging the bound.

- `cargo test --workspace --exclude bin-e2e --exclude plugingo-e2e` — green (~1900 tests)
- `plugingo-e2e` — solo-green (16 + 9 codegen serially). Under full parallel load its codegen tests ENOSPC on this host; that is the known environmental issue, and re-running them alone passes.
- `lint` — clean. `blocking::` suite 12× consecutively and the full `core` suite 4× to clear a flake (below).
- **Mutation-verified** (broken, confirmed red, reverted): lifting the semaphore overshoots the bound test; dropping the job marker fails the marker + witness tests; removing the LSP `enter()` panics with "there is no reactor running"; a non-String lost-job payload fails the lost-job test.

Two defects in the *tests* surfaced and are fixed here, both worth naming:

1. `a_job_can_run_a_nested_job` holds a run slot while acquiring a second — interleaved with a test that drains every slot, it deadlocked the whole `core` binary on the fair semaphore. It now takes the suite mutex, and the one drain-all that *parked* on the queue uses try-acquire + yield so it can never head-block a permit holder.
2. `trim_batch_recovers_a_guard_released_during_the_delay` is timed (the batch lost its last steppable hook along with the flushes), and a concurrent `lint` starved its probes so the guard was free by the first one. Margins widened 50ms/3s → 500ms/2s and re-verified under that same load.

One measured surprise, recorded in the code: **tokio 1.52 drains its blocking queue on shutdown**, so a job merely queued when a shutdown begins runs rather than being dropped. My first lost-job test assumed the opposite and failed; `run`'s non-panic `JoinError` branch is therefore the already-shut-down-runtime case, which the final test provokes deterministically.

## Review board

- **code-quality**: BLOCKED on the cross-test deadlock above (found independently) → fixed, stress-verified.
- **feature-quality**: BLOCKED on the LSP contract being untested → fixed, plus both MAJORs (lost-job branch, nested `run`) and three MINORs on permit/marker release paths.

## Bench gate pending

`heph-bench` was deliberately **not** run here — the serial bench session is coordinated separately, and a contended local run's noise floor would not be trustworthy. Performance parity for the mechanism switch was already measured in `docs/CONCURRENCY_MEASUREMENTS.md`; the Tier A/B comparison on this PR is still owed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -868,7 +868,6 @@ name = "core"
 version = "0.1.0"
 dependencies = [
  "anyhow",
- "crossbeam-channel",
  "enclose",
  "futures",
  "libc",
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -8,7 +8,6 @@ workspace = true
 
 [dependencies]
 anyhow = "1.0.102"
-crossbeam-channel = "0.5"
 serde = { version = "1", features = ["derive"] }
 tar = "0.4"
 futures = "0.3.32"
--- a/crates/core/src/blocking.rs
+++ b/crates/core/src/blocking.rs
@@ -1,375 +1,271 @@
-//! Fixed pool of long-lived OS threads for synchronous blocking work that must
-//! not run on a tokio runtime worker.
+//! Bounded [`tokio::task::spawn_blocking`] for synchronous work that must not
+//! run on a tokio runtime worker (tar/copy into the cache, gzip, borsh, sqlite
+//! spool reads, Starlark evaluation, package walks).
 //!
-//! Every other way of running blocking work is unusable here for a different
-//! reason:
+//! Running that work inline on a worker parks the worker with the runtime
+//! unaware — it neither hands the worker's queue off nor spawns a replacement,
+//! so enough concurrent jobs stop the reactor and the timer wheel and the
+//! build looks hung while nothing is deadlocked (the failure #180 fixed).
+//! [`run`] moves the job to tokio's blocking pool and awaits its
+//! `JoinHandle`, bounded by a semaphore.
 //!
-//! - **Inline on the worker** — what `hproc::process_supervisor::block_or_inline`
-//!   does on Linux. The runtime is never told, so it neither hands the worker's
-//!   queue off nor spawns a replacement: that thread simply stops polling. With
-//!   `worker_threads = ncpu` (2–4 on a CI runner) a handful of concurrent cache
-//!   writes park *every* worker, and then the reactor and the timer wheel stop
-//!   running too — in-flight HTTP transfers make no progress, their deadlines
-//!   never fire, the TUI freezes. Nothing is deadlocked and the build looks hung.
-//!   This is the failure this module exists to remove.
-//! - **`tokio::task::block_in_place`** — correct in principle (the runtime hands
-//!   off), but every call burns a worker handoff and pulls a fresh thread out of
-//!   the blocking pool, and it needs a runtime context this module cannot assume
-//!   (see below). A 0.94 → 0.74 concurrency regression was measured against it
-//!   in the #180 era; a 2026-07-31 re-test found parity within noise on macOS
-//!   Tier A (`docs/CONCURRENCY_MEASUREMENTS.md`) — treat the number as
-//!   historical, not as the reason this pool exists.
-//! - **`tokio::task::spawn_blocking`** — its `JoinHandle` wake-up rides tokio's
-//!   cross-thread waker, observed to drop wake-ups on macOS under heavy load (see
-//!   `docs/RCA_MACOS_WAKER.md` and the hazard note in `hproc::proc_exec`), which
-//!   strands the awaiting task.
+//! ## History
 //!
-//! So: a fixed set of named threads, a `crossbeam_channel` queue, and a
-//! `oneshot` for the result. The threads are created once and live for the
-//! process, so a job costs a channel send rather than a thread spawn.
+//! This module used to be a hand-rolled fixed pool of OS threads with a
+//! `crossbeam_channel` queue, plus a process-wide waker registry re-woken by a
+//! 250ms ticker thread ("the backstop"). Both existed because tokio's own
+//! path was unusable or distrusted:
 //!
-//! **Dropped-wake-up backstop.** The result still crosses threads, so [`run`]
-//! does not simply `await` the `oneshot` — a pending waiter is re-woken on a timer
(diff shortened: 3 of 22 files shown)


raphaelvigee:
## Bench gate — no measurable change on any scenario, and the instrument says how much to trust that

Tier A (in-process), interleaved rep-by-rep, macOS arm64 (M1 Pro, 10 core). Corpus = the CI perfbench shape (4000 targets / 200 pkgs / 10 layers / fan-out 4 / 60 go pkgs). Every binary in this table was built from the same tree with the same `--release` profile, so build-flag differences can't smear the result.

Baseline = `master` @ `4567cc09`, candidate = this PR @ `6c1e995f`.

| Pair | Scenario | Baseline (ms) | Candidate (ms) | Delta | ±1 SE |
|---|---|---:|---:|---:|---:|
| **this PR vs master** | full-hit (n=20) | 736.9 | 709.1 | −3.8% | ±7.8% |
| **this PR vs master** | incremental (n=20) | 752.4 | 728.0 | −3.2% | ±5.7% |
| **this PR vs master** | cold (n=6) | 82456 | 85514 | +3.7% | ±8.0% |
| control: master vs `d46c6c4f` | full-hit (n=20) | 522.8 | 534.2 | +2.2% | ±6.1% |
| control: master vs `d46c6c4f` | cold (n=6) | 91095 | 88672 | −2.7% | ±3.1% |
| **control: this PR vs itself** | full-hit (n=20) | 520.2 | 538.4 | +3.5% | ±5.1% |

### Verdict

**No regression, and no win either** — every delta is smaller than the same-binary control's own +3.5% offset. Correct outcome for a refactor whose product is −163 net lines rather than a speedup.

Read the last row before the others: two copies of the *identical* binary measured 3.5% apart. That is this session's noise floor, and it means this gate can honestly claim only **"no change larger than about ±8%"** — not "identical". The machine was busy (Chrome, `fseventsd`, and an MDM endpoint-security extension walking every file the bench writes; load average 125–190). Per-pair drift was worse still: the same master binary measured 522 ms in one pair and 737 ms in another, which is why only within-pair interleaved deltas are quoted and no number here is compared across rows.

The middle two rows exist because `master` moved past this PR's base while it was open (#309, #312 landed). They confirm that drift is itself inside noise, so it isn't hiding in the top three rows.

`cache_load` was not run — it is an instrument, not a gate, and reading it under this load would produce exactly the kind of number this comment is arguing against.

PR #322, opened 2026-08-02
test(bin-e2e): stop shell_pty waits matching their own keystrokes

raphaelvigee:
`exit_is_prompt_after_chatty_and_large_output` failed on `linux/arm64` ([job](https://github.com/hephbuild/heph/actions/runs/30758605612/job/91527491690)) with `never saw "shellpty#" within 10s`, while the command it was waiting on was still streaming output.

## The cause is not the 10s budget

A PTY echoes what we type, and `wait_for` only skips bytes *before* its cursor — the echo of the line we are about to send lands **after** it and matches like real output. So:

```rust
send_str("for i in $(seq 1 200); do echo err-$i >&2; done; echo stderr-batch-done\n");
wait_for("stderr-batch-done", 15s);   // matched our own keystrokes, instantly
wait_for("shellpty#", 10s);           // ...so this had to absorb all 200 stderr writes
```

The CI log shows it plainly — `shellpty# for i in ...; echo stderr-batch-done` came back as terminal echo and satisfied the wait before bash had run anything.

`wait_for`'s doc already names this hazard for the fixed `shellpty#` prompt and solves it with the cursor. The cursor cannot help when the needle is in the line being sent.

That is worse than an assertion that proves nothing: a wait that returns early hands the command's whole runtime to whatever waits **next**. Here that was the prompt sync, whose 10s then had to cover 200 stderr writes (or 900 KiB of output) the sentinel wait was supposed to have absorbed. On a loaded runner it did not.

## Fix

`echo_marker(head, tail)` builds the command and its marker together, splitting the literal on `''` so bash prints the marker whole while the echoed line cannot match it. Applied to all four self-matching sites.

Three of them were also **silently vacuous**: `marco-N-polo`, `still-alive` and `stderr-batch-done` were only ever observed as echoes — so "typed input reaches the child and its response comes back" and "the session is still usable after Ctrl-C" were never actually being tested.

The two waits in the chatty/large-output test move to `DEADLINE`. They are synchronization, not assertions: this test's claim is how fast `exit` returns, measured by `wait_exit`. A hang still fails, at 60s instead of 15s. The genuine timing assertions — Ctrl-C recovery, exit promptness — keep their tight bounds.

## Also: `e2e <file>` in CLAUDE.md was wrong

Hit while verifying this. Args are forwarded to `cargo test` verbatim, so the documented `e2e tui_pty` is a test-**name** filter. Nothing in `tui_pty.rs` is named `tui_pty`, so it matches nothing, runs **zero tests, and exits 0** — the same failure mode this PR is about, a green result that tested nothing. Selecting a file needs `--test`.

## Verification

`e2e --test shell_pty` against the real release binary and plugin cdylibs:

```
running 4 tests
test echo_marker_prints_the_marker_without_typing_it ... ok
test shell_echoes_typed_input_and_exits_promptly ... ok
test exit_is_prompt_after_chatty_and_large_output ... ok
test ctrl_c_interrupts_foreground_child_not_the_session ... ok
test result: ok. 4 passed; 0 failed
```

`echo_marker_prints_the_marker_without_typing_it` pins the helper in both directions, including that bash really folds the quotes — checked against 3.2.57, the bash these sessions get on macOS. It spawns no heph binary and needs no staged dist.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -31,11 +31,14 @@ The `gen` script is a devenv-provided alias, assume its present. It must be run 
 
 ```bash
 e2e                          # build the artifacts from this tree, then test them
-e2e tui_pty                  # one test file
+e2e --test tui_pty           # one test file
+e2e restores_the_terminal    # one test, by name substring
 e2e -- --nocapture           # args after `--` reach cargo test
 HEPH_E2E_FROM=dist e2e       # test an already-downloaded artifact set (what CI does)
 HEPH_E2E_KEEP_DIST=1 e2e     # keep the staged artifacts instead of deleting them
 ```
+
+Selecting a file needs `--test`. Everything after `e2e` is forwarded to `cargo test` verbatim, so a bare `e2e tui_pty` is a *test-name* filter — and no test in `tui_pty.rs` is named `tui_pty`, so it matches nothing, runs zero tests, and **exits 0**. A run that tested nothing is indistinguishable from a run that passed; always check the `running N tests` count.
 
 One script, one code path — CI runs the same `e2e`, differing only in where the artifacts come from. Do not add a parallel script or inline the steps into the workflow.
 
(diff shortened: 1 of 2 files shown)


raphaelvigee:

PR #323, opened 2026-08-02
fix(diag): stop SIGQUIT deadlocking the process it is diagnosing

raphaelvigee:
Sending `SIGQUIT` to a real run froze heph solid — every thread at 0% CPU, TUI dead, build unrecoverable. The tool built to diagnose hangs was causing one.

## Cause

`on_dump_signal` (`src/diag.rs`) runs **inside a signal handler** and calls:

- `Backtrace::force_capture()` — the DWARF unwinder, which takes libgcc's global `object_mutex` and re-enters `dl_iterate_phdr`
- `format!(…)` — which allocates

Neither is async-signal-safe. A thread interrupted inside `malloc` re-enters the allocator from its own handler, deadlocks on the arena lock it already holds, and never returns. Every other thread then blocks on that same lock.

The old comment justified this as *"a pragmatic trade"* on the grounds that it *"targets a CPU-bound hang, where the interrupted thread is looping in compute rather than inside the allocator"*. That premise doesn't hold for this engine: a CPU profile of a real run puts **~11% of its time in drop glue alone**, so at any instant some worker is very likely inside the allocator. With 12–16 threads signalled, hitting one is near-certain rather than unlucky.

Serialising the sweep never mitigated it either — the `MAX_THREADS` cap and the 2ms inter-thread gap address *two* threads unwinding at once, but the deadlock needs only **one** thread interrupted in the wrong place.

## Fix

The backtrace half moves behind `--diag-backtrace`, off by default.

`SIGQUIT` still works unconditionally and still writes the **in-flight report** — every open computation, how long it has been open, and what it is waiting on. That half is produced by an ordinary thread with no re-entrancy hazard, and it is the one this file already documents as mattering:

> Thread backtraces alone answered "every thread is idle" on a wedged run — true, and useless. The inventory is the half that names the stuck work.

The `warn!` line names the flag, so a reader who wanted stacks and got none learns they exist rather than concluding the dump is broken.

This is the same defect class as #318, which took the SIGPROF sampler off `_Unwind_Backtrace` for exactly this reason. Walking frame pointers here too would make the backtrace half *safe* rather than merely gated — worth doing, but a much larger change; until then the flag states what it costs.

## Test

`a_dump_installs_no_handler_unless_backtraces_were_asked_for` asserts on `SIGUSR1`'s **disposition** after a default dump, not on the absence of backtrace text — an empty sweep would satisfy the latter, while a handler still installed is the actual footgun. It also asserts the in-flight report is still written, since the point of the gate is to keep that half, not to make `SIGQUIT` do nothing.

Only the *off* path runs end to end: driving the on path in-process would signal every thread of the test binary — i.e. reproduce the deadlock. The flag round trip is asserted inside that same test because `BACKTRACES` is process-global, and a parallel sibling test setting it would be that reproduction by accident.

## Compatibility

Additive CLI flag, default `false`. The only behaviour change to an existing invocation is that `SIGQUIT` no longer emits thread backtraces unless asked — which is the fix.

### DIFF
--- a/crates/core/src/hmemoizer/mod.rs
+++ b/crates/core/src/hmemoizer/mod.rs
@@ -8,7 +8,7 @@ use std::future::Future;
 use std::hash::{Hash, Hasher};
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
-use std::sync::{Mutex, OnceLock, PoisonError};
+use std::sync::{Mutex, OnceLock};
 use std::time::Duration;
 use xxhash_rust::xxh3::Xxh3Default;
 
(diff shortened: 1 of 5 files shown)


raphaelvigee:

PR #324, opened 2026-08-03
perf(memoizer): poll a cold body inline, spawn only if it suspends

raphaelvigee:
Stops handing tokio a task for every memoized computation that never suspends. **4.6× wall at 6 cores, 10× at 10 cores** on a 192k-target warm resolution.

## The problem

`process` handed every cold cell to a tokio task. That's the right model for a computation that *suspends* — cancellation is an `abort()`, the body is always polled by the runtime. But most memoized computations resolve without ever yielding, and for those the task is pure overhead:

- `OwnedTasks` push + a global `added` atomic on spawn
- a shard-mutex unlink + a global `count` atomic on completion
- the `Notify` wake machinery in between

Those two counters are **process-global, touched by every worker**, so the cost is contention rather than work. Profiling 192k targets warm — identical work, only the core count differing:

| | 1 core | 10 cores |
|---|---|---|
| `tokio OwnedTasks::remove` | 3.82s CPU | **25.88s CPU** |
| `notify_waiters` | 2.86s | 9.60s |
| `wake` / `wake_by_ref` / `arc_wake` | 2.58s | 10.92s |

tokio's task machinery was **20.9%** of the profile, and `OwnedTasks::remove` its single largest flat line.

It also made the engine scale *backwards*: the old path is **slower at 10 cores (248.5s) than at 4 (202.9s)**.

## The change

Insert the cell, **release the cache lock**, poll the body once on this stack, spawn only on `Pending`.

Releasing the lock first is mandatory — a body that completes inline calls `publish`, which takes that same lock. It's safe because the caller already holds an interest, so no concurrent `cancel_abandoned` can evict the cell mid-poll.

Polling with a no-op waker is sound: a future must honour the waker from its most recent poll, and tokio polls immediately on spawn, so anything registered against the no-op waker is re-registered against the real one before it could be woken.

## Measured

Same binary, `HEPH_MEMOIZER_INLINE` the only difference, 192k targets warm. **The win is strongly core-count dependent** — reporting one number would be misleading:

| cores | spawn wall | inline wall | speedup |
|---|---|---|---|
| 1 | — | 321.3s | — |
| 2 | — | 283.1s | — |
| 4 | 202.9s | 153.3s | 1.3× |
| **6** | **262.9s** | **56.8s** | **4.6×** |
| 10 | 248.5s | 24.8s | **10.0×** |

CPU at 6 cores: 578.7s → 142.0s (4.1×). At 10 cores: 706.4s → 87.7s (8.0×).

Peak RSS drops too — a task per cell is a task *harness* per cell, plus the `Notify` and `JoinHandle` state hanging off it:

| mode | peak RSS |
|---|---|
| spawn | 13.6 GB |
| inline | **7.7 GB** |

**43% less memory** for the same 192k-target resolution (~70 KB/target → ~40 KB/target).

Both modes report `done 58131, err 0, cache 19553 hit / 0 miss` — identical work, not a short-circuit.

The old path *degrades* with more cores; the new one scales superlinearly (11.5× from 1→10 cores), because the contention it removes grows with core count. On 1–2 cores there is no win, which is expected: there is no contention to remove.

## Stack depth

The second commit fixes a stack overflow CI caught on all three targets (`deep_warm_chain_completes_on_a_2mib_stack`).

Spawning per cell was buying something beyond isolation that nothing named until it was gone: a spawned body starts on its **own** stack, so a memoized descent cost O(1) stack per level however deep the graph. Inline recursion puts it all on the caller's stack. Fixed with `stacker::maybe_grow`, the same mechanism and constants `engine::grow_stack` already uses for the transparent-group re-inline. `grow_stack`'s doc is corrected rather than left stale.

A recursion-depth cap was also tried as an alternative and **measured no better than the stacker wrapper alone** (within noise at every core count), so it is not included.

## What does and doesn't change

**Unchanged:** a body that suspends still gets a task and is still abortable; the handle is stored under the cache lock before the body can publish, so a `Done` cell never holds a live handle.

**Changed:** a body that never suspends runs to completion on the caller's stack and cannot be cancelled — not a loss, since there is nothing to cancel once it has finished.

`TaskCell::inline` keeps `task_live()` truthful during the inline window, so a `SIGQUIT` dump taken mid-poll doesn't report the cell as stranded.

## Tests

Four existing tests needed their **setup** fixed, not their assertions weakened: each aborts or spawn-fails against a body that no longer needs a task, so each now yields once first — the harness constructing the scenario it always meant to.

Two new tests freeze both halves:
- `a_body_that_never_suspends_is_driven_inline` — completes on the first poll, holds no task handle
- `a_body_that_suspends_still_gets_a_task` — guards against regressing to "never spawn"

The first deliberately **fails** under `HEPH_MEMOIZER_INLINE=0` rather than silently skipping.

Engine suite 508 passed, core 125 passed, workspace clippy clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -873,6 +873,7 @@ dependencies = [
  "libc",
  "rustc-hash",
  "serde",
+ "stacker",
  "tar",
  "tempfile",
  "tokio",
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -14,6 +14,7 @@ futures = "0.3.32"
 enclose = "1"
 xxhash-rust = { version = "0.8.15", features = ["xxh3", "std"] }
 rustc-hash = "2"
+stacker = "0.1"
 libc = "0.2"
 tokio = { version = "1.52.3", features = ["rt", "rt-multi-thread", "macros", "sync", "time", "test-util"] }
 tracing = "0.1"
--- a/crates/core/src/hmemoizer/mod.rs
+++ b/crates/core/src/hmemoizer/mod.rs
@@ -1710,9 +1710,15 @@ mod tests {
             .await
             .expect("shut down the runtime");
 
+        // Suspends, so finishing it requires a spawn — which is the thing that
+        // fails on a dead runtime. A body that completed inline would need no
+        // runtime and would legitimately succeed (see `task_cell::inline_first`).
         let v = tokio::time::timeout(
             Duration::from_secs(5),
-            m.once("k".to_string(), || async { Ok(Arc::new(1u32)) }),
+            m.once("k".to_string(), || async {
+                tokio::task::yield_now().await;
+                Ok(Arc::new(1u32))
+            }),
         )
         .await
         .expect("a dead runtime must fail fast, not hang");
--- a/crates/core/src/hmemoizer/task_cell.rs
+++ b/crates/core/src/hmemoizer/task_cell.rs
@@ -172,6 +172,14 @@ impl<V> TaskCell<V> {
     }
 }
 
+/// If less than this much stack remains when a body is polled inline, grow.
+/// Matches `engine::grow_stack`'s red zone — sized for the ~100 KiB frames a
+/// memoized descent puts on the stack.
+const INLINE_RED_ZONE: usize = 512 * 1024;
+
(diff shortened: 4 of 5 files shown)


raphaelvigee:

Issue #326, opened 2026-08-03
perf: engine was tokio-task-registry contention bound; what is fixed and what is left


raphaelvigee:
Investigation into why heph used only ~30% of its CPU budget on a large resolution-heavy workload, what fixed it, and what is left. Written so another session can pick this up without re-deriving anything.

## TL;DR

The engine was **contention-bound in tokio's task registry**, not CPU-bound. #324 fixes it: **4.6× wall / 4.1× CPU at 6 cores, ~10× at 10 cores, 43% less peak RSS**. After that the workload is **depth-bound**, and the remaining CPU slices (each 2–5% of real CPU) are believed worth pursuing — see [Remaining work](#remaining-work).

---

## 1. What was wrong

Production symptom: a 6-minute run used **1.78 of 6 cores (30%)** of a `limits.cpu: 6` pod.

Ruled out, each with a measurement:

| hypothesis | verdict | evidence |
|---|---|---|
| CFS throttling | no | `nr_throttled` 0.91% of periods over pod lifetime |
| thread oversubscription | no | `taskset -c 0-5` changed nothing |
| `dep_dag` mutex convoy | no | stall report: 0 lock waits, 0 in-flight cells, 12/12 workers free |
| semaphore saturation | no | no `limits` line in stall report |
| CPU hotspots | no | removing 33s of hashing (#319/#320) moved wall time **zero** |

Actual cause: **`tokio::runtime::task::list::OwnedTasks`**. Every spawn does `added.add()` + a shard push; every completion does a shard-mutex unlink + `count.decrement()`. `added`/`count` are **process-global atomics touched by every worker**, so cost scales with contention, not work.

Identical work, 192k targets, only core count differing:

| | 1 core | 10 cores |
|---|---|---|
| `OwnedTasks::remove` | 3.82s CPU | **25.88s CPU** |
| `notify_waiters` | 2.86s | 9.60s |
| `wake`/`wake_by_ref`/`arc_wake` | 2.58s | 10.92s |
| **tokio task machinery total** | — | **20.9% of profile** |

The engine scaled *backwards*: **slower at 10 cores (248.5s) than at 4 (202.9s)**.

Not `tokio_unstable`-specific — checked; `cfg_64bit_metrics` is only `target_has_atomic`, and `MetricAtomicUsize::decrement` is unconditional.

## 2. The fix (#324)

[...]

PR #327, opened 2026-08-03
fix(pprof): keep the samples taken inside libc, stop unwinding them

raphaelvigee:
`--pprof-cpu` was deleting roughly half of every profile and saying nothing about it.

`UNWIND_BLOCKLIST` named the C library, and pprof applies that list **twice** — to the *interrupted PC* as well as to each walked frame. A sample whose leaf landed in `malloc`, `free`, `memcpy` or a syscall stub was discarded whole, before anything was walked. heph is in libc constantly, so what came out was a profile of everything except the allocator, the copies, and the syscalls.

## How much was missing

Measured on a warm 85k-target `validate //go/large/...` with samply, which has no such filter:

| slice | share of on-CPU samples |
|---|---|
| platform allocator (`libsystem_malloc`) | ~15% |
| `_platform_memmove` / `memset` / `memcmp` | ~10% |
| FS syscalls (`stat`, `getattrlist`, `open`, `getxattr`) | ~9% |

`--pprof-cpu` reported **none** of it. That is how ["removing 33s of hashing moved wall time zero"](https://github.com/hephbuild/heph/issues/326) could be true while the profile looked like hashing was the problem.

## Why keeping them is free

The frame-pointer walk never starts inside the interrupted function. It starts at the return address in the innermost frame record — which for a sample taken in `malloc` is **heph's own code**. It climbs heph frames from there and never enters libc at all. This is the same one-frame shift already documented on `start` (the one that makes `flat` read as "time in the callee").

So the entry bought nothing but the deletion. What stays on the list is what the walk must not *climb into*: the unwinder (`libgcc_s`, `libunwind`), the loader (`ld-linux`, `libdyld`, `vdso`), and GCD.

## The test

The new `libc_only` half of the sampling test is a leaf that sits inside a 1 MiB `memmove` for its whole duration — the shape the blocklist used to erase. Verified in both directions on darwin/arm64:

| | stacks naming `libc_only` | stacks naming `cpu_only` |
|---|---|---|
| C library **off** the list (this PR) | > 0 | 21 |
| C library **on** the list (before) | **0** | 21 |

`cpu_only` stays green either way. That asymmetry is exactly how the missing half went unnoticed for so long.

## One caveat, documented rather than `cfg`-split

The walk reads whatever frame-pointer register the interrupted function left behind. Both aarch64 targets reserve `x29` for the frame record, so the stack is real. glibc on x86_64 is built `-fomit-frame-pointer` and `%rbp` is only callee-saved, so a libc routine *using* `%rbp` as scratch when `SIGPROF` lands yields a fabricated stack — never a fault, since every frame address is probed before it is dereferenced.

All three targets keep the sample and attribute it to the caller; only x86_64 can mis-name some of them. Blocklisting libc on x86_64 alone would trade that back, and a profiler that silently omits half its subject is the worse failure of the two. Say the word if you'd rather have the arch split.

Stacked on #324.

### DIFF
--- a/src/pprof_dump.rs
+++ b/src/pprof_dump.rs
@@ -37,42 +37,68 @@ const SAMPLE_HZ: libc::c_int = 199;
 
 /// Libraries the sampler must not walk into.
 ///
-/// `pprof` resolves these names against the loaded shared objects, checks the
-/// interrupted PC *before* walking anything, and — because this build enables
-/// the `frame-pointer` tracer (see [`start`]) — checks every frame it reaches
-/// afterwards. A sample whose leaf lands in one of them is dropped whole; a walk
-/// that climbs into one stops there. That costs only frames attributable to the C
-/// library, which carry no information about heph's own hot loops.
+/// `pprof` resolves these names against the loaded shared objects and applies
+/// them twice: it drops a whole sample whose *interrupted PC* lands in one, and —
+/// because this build enables the `frame-pointer` tracer (see [`start`]) — it
+/// ends the walk at the first frame that reaches one. Both checks read this one
+/// list, so an entry here buys frame-level containment at the price of every
+/// sample that happened to be interrupted inside that library.
 ///
 /// This used to be the *entire* defence, back when the tracer was
 /// `_Unwind_Backtrace` and only the leaf PC could be tested — see [`start`] for
 /// why that was never sound and what replaced it.
 ///
+/// **The C library is deliberately not here**, and that is the difference
+/// between a profile and a fiction. heph is in libc constantly — `malloc`,
+/// `free`, `memcpy`, every syscall stub — so blocklisting it dropped roughly
+/// half of every profile with nothing in the output to say so: a warm 85k-target
+/// resolution spends ~15% of its CPU inside the allocator and ~10% in `memmove`,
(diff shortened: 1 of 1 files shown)


raphaelvigee:
`Binary E2E darwin/arm64` is red here, and it is **not this PR's**. It is the pre-existing `exit_is_prompt_after_chatty_and_large_output` PTY flake, inherited from the base.

Same test, same run, rotating target:

| PR | darwin/arm64 | linux/amd64 | linux/arm64 |
|---|---|---|---|
| #322 | pass | **fail** | pass |
| #323 | **fail** | pass | pass |
| #324 | **fail** | pass | **fail** |
| #327 (here) | **fail** | pass | pass |
| #330 (this + 2 more commits) | pass | pass | pass |

Every failure is `never saw "large-output-done" within 60s` at `crates/bin-e2e/tests/shell_pty.rs:221`. Which target goes red changes run to run, and #330 — which contains this commit plus mimalloc plus the memoizer sharding — is green on all three, so nothing here causes it.

#322 is already the fix in flight for exactly this wait (`test(bin-e2e): stop the large-output wait timing out`), so it belongs there, not in this PR.

Test / Lint / Build are green on all three targets here.

raphaelvigee:

PR #362, opened 2026-08-05
feat(plugin-js): oxc-based import-graph resolver, phantom-dep detection

raphaelvigee:
M2 of the JS/TS heph plugin plan (part 2/6 of the stack).

Fixes the deferred optionalDependencies hard-fail (platform-restricted optional deps now silently skip instead of failing Provider::get), then adds a real import-graph resolver on top of M1's package.json-declaration wiring.

- oxc_parser + oxc_resolver extract and resolve import/require/dynamic-import specifiers per real Node condition-set semantics (separate ESM/CJS/types resolvers), building two distinct graphs (runtime vs type-only edges).
- Phantom-dependency detection: an import resolving into a package not in the declared-dependency closure (deps + devDeps + peerDeps) is a hard error naming the file/specifier/package, hermetic against a fresh checkout (no ambient node_modules required) via a bare-specifier name check.
- Conformance corpus covering exports-map condition ordering, wildcard specificity, array fallbacks, null-blocked subpaths, self-referencing imports, and the "imports" field — cross-checked live against a real Node binary when present, self-gated otherwise.

Reviewed by feature-quality/code-quality/hermeticity. Four BLOCKERs found and fixed: the phantom-dep check being a no-op without ambient node_modules, peerDependencies never counted as declared, an oxc_resolver default reading the ambient NODE_PATH env var, and a resolved-but-unclassifiable edge silently passing instead of failing closed. Each has a regression test.

Explicitly deferred: import-equals require() and require.resolve() extraction, per-specifier `type` modifier detection, provider-lifetime resolver caching (fixed later in the stack), pinning a hermetic Node toolchain for the conformance corpus's live cross-check.

## Test plan
- [x] cargo test -p plugin-js

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -66,7 +66,7 @@ dependencies = [
  "bumpalo",
  "ctor",
  "hashbrown 0.16.1",
- "num-bigint",
+ "num-bigint 0.4.8",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


raphaelvigee:

PR #363, opened 2026-08-05
feat(plugin-js): js_typecheck driver, per-package tsc

raphaelvigee:
M3 of the JS/TS heph plugin plan (part 3/6 of the stack).

A cacheable js_typecheck ManagedDriver running `tsc --noEmit` per package, fed by the M2 import graph rather than blind package.json declarations.

- Toolchain: tstool=host (only supported mode) resolves tsc from node_modules/.bin or PATH, queried once per Provider lifetime and hashed alongside the tsconfig content — disclosed non-hermetic escape hatch.
- Input scoping fixed through review to actually match what tsc reads: plain (non-type-only) cross-package imports, not just `import type`; third-party .d.ts inputs resolved via the same lockfile mechanism as js_install; tsconfig include/exclude honored; the full tsconfig extends chain declared and hashed; a shared/ancestor tsconfig with unscoped include is now a loud Provider::get error instead of a silently unsound cache key.

Reviewed by feature-quality/code-quality/hermeticity — five BLOCKERs found (all variations on "the declared Input set doesn't match what tsc actually reads") and fixed, each with a regression test proving the specific divergence. Tests requiring a real tsc binary are #[ignore]d with a named reason rather than silently skipping and reading as a pass.

## Test plan
- [x] cargo test -p plugin-js


raphaelvigee:

PR #364, opened 2026-08-05
feat(plugin-js): js_test driver, per-test-file granularity

raphaelvigee:
M4 of the JS/TS heph plugin plan (part 4/6 of the stack).

A cacheable js_test ManagedDriver with one target per test file (vitest default, jest alt via a single testrunner config option), fed by the M2/M3 import graph — the differentiator over Turborepo/Nx, which never get finer than per-package task caching.

- Per-test-file Input scoping: the test file's own runtime-transitive closure (BFS over ImportGraph::runtime_edges, bounded to the owning package), third-party deps resolved via the same lockfile-driven mechanism js_install/js_typecheck use, and the resolved test-runner config.
- Runner-config discovery covers vitest's own fallback (vite.config.*) and jest's package.json "jest" field, not just the dedicated config filenames.

Reviewed by feature-quality/code-quality/hermeticity. Three BLOCKERs found and fixed: **a real sandbox escape** — a js_test addr's `file=` argument accepted an absolute path or a `..`-escape with no validation, letting a target read/exec outside the workspace and sandbox entirely (fixed at both Provider::get and defensively again in the driver's run()); vitest/jest config-referenced files (setupFiles, globalSetup, a shared base config) were untracked, so editing one didn't bust the cache of every dependent test; and the primary runner-config file itself could go undiscovered for vite.config.\*/package.json-jest-field layouts. Each has a regression test.

Explicitly deferred: per-test-file Provider::get rebuilding the whole package's import graph from scratch (fixed later in the stack); the cross-package one-hop trim risking a stale test-pass cache hit for a barrel-file re-export; test-runner-specific alias resolution.

## Test plan
- [x] cargo test -p plugin-js

### DIFF
--- a/crates/plugin-js-cdylib/src/lib.rs
+++ b/crates/plugin-js-cdylib/src/lib.rs
@@ -1,24 +1,29 @@
 //! The JS/TS plugin as a loadable cdylib behind the stable ABI.
 //!
-//! **M0-M3 scope**: exports the `js` provider (package discovery + pnpm/npm
+//! **M0-M4 scope**: exports the `js` provider (package discovery + pnpm/npm
 //! workspace-member resolution + lockfile-driven dependency wiring + oxc-based
 //! import-graph resolution), its `js_package_info` dependency-wiring driver,
-//! the hermetic `js_install` third-party fetch driver, and the `js_typecheck`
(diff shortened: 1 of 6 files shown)


raphaelvigee:

PR #367, opened 2026-08-05
feat(bench): add JS/TS Tier B scenarios, go/js/both selector

raphaelvigee:
Part 7/7 of the stack — extends `heph-bench` to measure the new plugin-js the same way it already measures plugin-go, generalizes Tier B (`dist.rs`) across languages instead of duplicating the go-specific path, and publishes the js plugin as a real, generally-installable artifact — same as go/gha.

## What's here

- **`crates/bench-corpus`**: `generate_js_tree` produces a synthetic, network-free pnpm workspace (zero third-party deps by design — no js_install fetch, no lockfile, fully hermetic) with the same layered-DAG shape the bash and go generators already use. Wired via `js_packages`/`js_max_depth`, mirroring `go_packages`/`go_max_depth` exactly — `0` (the default) is a no-op, so existing callers get byte-identical output.
- **`crates/bench`**: `dist.rs` rewritten around a `Lang` abstraction (GO/JS consts: name, provider-option fragment, package-count accessor, incrementalize fn) instead of two near-duplicate go/js code paths. `run dist --lang go|js|both` (default `go`, preserving existing-caller behavior exactly — byte-identical `.hephconfig`/scenario names to the pre-existing code path). `both` reports each language as its own distinct `ScenarioResult` (`<scenario>-go`/`<scenario>-js`) — never silently summed/averaged.
- The `corpus` CLI subcommand was missing `--js-packages`/`--js-max-depth` entirely (bench-corpus's library supported them, main.rs never exposed them) — added, mirroring `--go-packages`/`--go-max-depth`.

## A real pre-existing bug, found while mirroring the go path for js

Tier B's `heph r build //<lang>/...` invocation is the two-positional-arg form, which parses as `label("build") && //<lang>/...` — no BUILD file in the corpus sets that label, so it has **always matched zero targets and exited 0, silently measuring an empty build**. This affected the existing go benchmark too, not just the new js one — it predates this PR entirely, just surfaced while faithfully mirroring the path. Fixed to the query form (`heph r -e '//<lang>/...'`), verified against a real binary (built a fixture, confirmed the old form reports \"matched 0 targets\" and the new one actually executes targets), with a regression test that fails if the matcher regresses back to the broken form.

## Artifact pipeline — js is now published like go/gha

`plugin-js-cdylib` is built and published by `heph.yml`'s `build` job alongside `plugin-go-cdylib`/`plugin-gha-cdylib`, in the same pre-release artifact bundle, **and** `upload_artifacts` now generates a public `heph-js-plugin.json` manifest via `tools/pluginmanifest`, exactly the same way go/gha do (per-os/arch checksum entries, URL-base pointing at the release, a `.sha256` sidecar for the manifest itself). A real workspace can now opt in the same way as go/gha:

```yaml
plugins:
  - path: .heph3/heph-js-plugin.json   # or a url: pointing at the published release asset
```

`crates/bench/src/dist.rs`'s `write_dist_config` is unaffected — it still builds its own local manifest straight from the downloaded raw dylib and never reads the published one; the two are independent consumers of the same underlying cdylib asset.

`devenv.nix`'s `e2e` script still deliberately does **not** build `plugin-js-cdylib` locally — there is no `shipped_js_cdylib_loads` bin-e2e test yet to justify the build cost, mirroring the existing lint-but-don't-e2e-test asymmetry already in place for this crate. That's a narrower, still-open gap than manifest publishing and unaffected by this update.

## Known, disclosed gap

No `oxlint`/`esbuild` toolchain is provisioned on any CI runner yet, so `perf.yml`'s new JS Tier B comparison step will report \"no results\" until one is. The target-matcher fix above makes that failure mode honest (a real attempt, a real failure) rather than the previous silent zero-targets false success. Provisioning a JS toolchain for CI (host vs. hermetic, which versions) is a separate follow-up decision, not made here.

## Test plan
- [x] `cargo build -p bench -p bench-corpus`
- [x] `cargo test -p bench -p bench-corpus` (14 + 8 passed, 2 pre-existing network-gated ignored)
- [x] `cargo clippy -p bench -p bench-corpus --all-targets -- -D warnings` (clean)
- [x] `cargo fmt --check -p bench -p bench-corpus`
- [x] `cargo build --release -p plugin-js-cdylib` (the artifact this whole change assumes exists)
- [x] Workflow YAML (`heph.yml`, `perf.yml`) parses
- [ ] CI (`tst`, `lint`) on this PR

### DIFF
--- a/.github/workflows/heph.yml
+++ b/.github/workflows/heph.yml
@@ -260,6 +260,7 @@ jobs:
           PLUGIN_GO_NAME="heph-go-plugin_${{ matrix.os }}_${{ matrix.arch }}.$ext"
           PLUGIN_GHA_NAME="heph-gha-plugin_${{ matrix.os }}_${{ matrix.arch }}.$ext"
           PLUGIN_OCI_NAME="heph-oci-plugin_${{ matrix.os }}_${{ matrix.arch }}.$ext"
+          PLUGIN_JS_NAME="heph-js-plugin_${{ matrix.os }}_${{ matrix.arch }}.$ext"
           # heph-bench: the perf-regression harness (`perfbench` job). Publishing
           # it as a release asset alongside `heph` itself means that job never
           # builds anything to compare a PR against its baseline — both the
@@ -277,6 +278,7 @@ jobs:
           echo "plugin_go_name=$PLUGIN_GO_NAME" >> $GITHUB_OUTPUT
           echo "plugin_gha_name=$PLUGIN_GHA_NAME" >> $GITHUB_OUTPUT
           echo "plugin_oci_name=$PLUGIN_OCI_NAME" >> $GITHUB_OUTPUT
+          echo "plugin_js_name=$PLUGIN_JS_NAME" >> $GITHUB_OUTPUT
           echo "bench_bin_name=$BENCH_BIN_NAME" >> $GITHUB_OUTPUT
           echo "debug_bin_name=$DEBUG_BIN_NAME" >> $GITHUB_OUTPUT
           OUT="$CARGO_TARGET_DIR/${{ matrix.target }}/release"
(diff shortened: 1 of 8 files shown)


raphaelvigee:

PR #381, opened 2026-08-06
feat(plugin-oci): oci_load tags every image by content

raphaelvigee:
Based on #159 (`oci_load` only exists there). Independent of #378 — a sibling, not a link in that stack.

## What

`tag` was required. Now every load also applies a **content-addressed** tag — the image target's address as a repository, this load's input hash as the tag:

```console
$ heph run //app:load
app/img:9f2c4e1b7a0d3856
```

`tag` becomes optional and is applied **in addition**, never instead.

## Why

A tag written in a BUILD file is a moving target. It names whichever image was loaded *last*, so two branches, two people, or a rebuild after a change all collide on it and `docker run app:dev` silently runs the wrong thing. It is also unusable from a script, which has to read the log to find out what happened.

The derived tag can't do that: it changes when and only when the image changes, two people on the same commit get the same string, and it can be **predicted from the graph** instead of discovered after the fact.

The pinned tag is applied first, so an image is never briefly reachable only under a moving name.

## The repository follows the image, not the load

`//app:load` loading `//svc/api:img` gives `svc/api/img`, not `app/load`. The repository names what you are about to run, and nobody names their load target after the image it loads.

## Deriving it is not a formatting exercise

A docker reference is much narrower than a heph address — each path component is `[a-z0-9]+([._]|__|[-]+[a-z0-9]+)*`:

| Address | Repository | Why |
|---|---|---|
| `//cmd/server:img` | `cmd/server/img` | |
| `//Cmd/My_Server:Img` | `cmd/my-server/img` | grammar is lowercase alphanumeric |
| `//example.com/svc:img` | `example-com/svc/img` | **docker reads a first component containing a dot as a registry host** — kept as-is, `docker run` becomes a network pull from a registry that does not exist |
| `//localhost:img` | `heph/localhost/img` | a host by name alone, same trap |
| `//app:img@v=linux_amd64` | `app/img-v-linux-amd64` | variants are different images; sharing a repository puts two rows in `docker images` with nothing to tell them apart |
| `//app:___` | error at parse | names the way out (`give an explicit tag`) rather than a reference-format complaint from the daemon later |

The repository is resolved in `parse`. `hashin` is only known at run time — the engine computes it after `parse` — so only the repository is in the def.

`OCI_LOAD_FORMAT_VERSION` 3 → 4.

## Tests

9 unit tests covering the mapping table above, plus `tag` being genuinely optional and the repo tracking the image rather than the load. The daemon-gated test now asserts **both** tags land in the daemon and that the derived one is a hex hash — with a per-process package so its repository is unique, since the daemon is shared state. That assertion needs a real daemon: the archive carries `RepoTags`, but a containerd-backed daemon ignores them in favour of the OCI path, so the explicit tagging afterwards is invisible to a fake.

Drive-by: `test_real_skopeo_loads_an_oci_archive_into_the_daemon` had described skopeo since skopeo stopped being involved. Renamed and re-documented to what it actually covers.

https://claude.ai/code/session_01WqQZxsTqanJRWLWdPSchBo

### DIFF
--- a/crates/e2e/tests/oci_docker.rs
+++ b/crates/e2e/tests/oci_docker.rs
@@ -630,7 +630,9 @@ async fn test_real_docker_load_tags_the_image_in_the_daemon() -> anyhow::Result<
     // The package is unique per process so the *derived* repository is too:
     // parallel runs, and leftovers from a killed one, share the daemon.
     let pkg = format!("app{}", std::process::id());
-    let repo = format!("{pkg}/img");
+    // One `_`-joined component, matching `docker_repo`: a `/` path would let
+    // docker read the first segment as a registry host.
+    let repo = format!("{pkg}_img");
     let tag = format!("heph-e2e-oci-load:{}", std::process::id());
 
     let ws = workspace();
@@ -666,12 +668,16 @@ target(name = "autoload", driver = "oci_load", image = ":img")
     let explicit = Command::new("docker")
         .args(["image", "inspect", &tag])
         .output();
+    // Looked up by the package prefix rather than the exact repository, so a
+    // change to how the name is derived fails below with the name it actually
+    // used — not here with an empty list and nothing to go on.
     let derived = Command::new("docker")
         .args([
             "images",
             "--format",
             "{{.Repository}}:{{.Tag}}",
-            repo.as_str(),
+            "--filter",
+            &format!("reference={pkg}*"),
         ])
         .output()
         .map(|o| String::from_utf8_lossy(&o.stdout).trim().to_string())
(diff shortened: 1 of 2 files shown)


raphaelvigee:
Reshaped per review, and rebased on current `master`.

**One `_`-joined component, not a `/` path.** `//cmd/server:img` → `cmd_server_img`.

That's not cosmetic. A multi-component name invites docker to read the first one as a **registry host** — it does exactly that whenever the component holds a `.` or a `:` — so `example.com/svc/img` turns `docker run` into a network pull from a registry that does not exist. One component can never be read that way, and `docker images` gets one flat row per target. The `localhost` guard shrank to the only case left: a root-package target literally called `localhost`.

`_` is only ever the separator, one at a time between non-empty segments, because every character outside `[a-z0-9]` in an address segment already became `-`. That matters — the grammar allows `_` and `__` between alphanumerics but not `___` — so there's a test for it.

| Address | Repository |
|---|---|
| `//cmd/server:img` | `cmd_server_img` |
| `//Cmd/My_Server:Img` | `cmd_my-server_img` |
| `//example.com/svc:img` | `example-com_svc_img` |
| `//app:img@v=linux_amd64` | `app_img-v-linux-amd64` |
| `//:localhost` | `heph_localhost` |
| `//app:___` | error at parse |

**The derived name is now the default, not an addition.** An explicit `tag` replaces it instead of sitting beside it — which is what "if tag is not specified" asks for. The daemon-gated test grew a second, untagged `oci_load` target so both paths are covered.

Rebased onto `master` (`7593d4ea`), through #159. `cargo check --workspace --all-targets`, `lint` and the plugin + e2e suites are green on the rebased tree.

raphaelvigee:
Re-based to sit on top of #378 rather than beside it, so all three are one linear stack (#379): **#159 → #378 → #381**.

Nothing in this PR changed — its diff against #378 is still exactly `load.rs` + `oci_docker.rs`, and it touches no file #378 touches. It moved because `gh stack` can only append to the top of an existing stack, and #159 ← #378 was already registered.

The order reads better this way anyway: images assembled without docker, then loaded under a name derived from their content.

PR #382, opened 2026-08-06
feat(group): filter and relocate outputs without copying them

raphaelvigee:
Lets a target re-export another target's outputs at different paths, without copying them.

```python
target(
    name = "dist",
    driver = "group",
    deps = ["//app:bin", "//web:assets"],
    include = ["**/*.so", "bin/**"],
    exclude = ["**/*_test"],
    strip_prefix = "build/out",
    prefix = "release",
)

# and one-off, no target needed:
deps = {"web": "//web:assets[**/*.css]"}
```

## Why it's cheap

Relocating today means an exec target running `cp` — a sandbox, a subprocess, and a second full copy of every byte in **both** the local and the remote cache. This costs a few string operations.

A transformed group produces `ContentView` artifacts: zero-copy windows onto its deps' artifacts that rewrite entry paths as they stream out of `walk()` and forward each file's read handle untouched. It declares itself uncacheable, so no revision, blob, or remote push is ever created for it. The deps stay cached exactly as before, and the group is re-derived from a path list each build.

An earlier design stored a manifest referencing the source blob. That was dropped: cache blobs are keyed by `(addr, hashin, name)` rather than globally content-addressed, so a cross-revision reference would have pulled GC (~2k lines) and the remote cache (~3.8k lines) into it. Resolving the view at the engine layer instead gets the same zero-copy property with no cache-format, GC, or remote-cache change.

## The cache-correctness constraint

`hashin = H(driver, own_def.hash, sorted(dep hashouts))` (`crates/engine/src/engine/meta.rs:67`). Per-input metadata is not folded in.

A `transparent` target is expanded away *before* `hashin` is computed, so its config never reaches a consumer's cache key — membership changes are caught only because they change which dep hashouts fold in, and **a path transform changes no dep's hashout**. A transparent relocating group would therefore be invisible to the cache, and consumers would happily reuse entries built against the pre-relocation layout.

That is why a group with a transform becomes a real target. Its transform reaches consumers through its own def hash and its artifacts' derived hashouts.

**A group with no transform is untouched** — still transparent, still inlined, and its def hash is pinned byte-identical to the pre-feature digest by a test, so nothing cached is invalidated by this landing.

## `rename` has two forms

```python
rename = "bin/server"                       # the sole selected output
rename = {"app/build/out/x": "lib/x"}       # exact emitted paths
```

The string form names no source path at all — it renames whatever survives `include`/`exclude`, which must be exactly one file — so it keeps working when a dep changes where it writes, or moves package. `include`/`exclude`/`strip_prefix` likewise accept the spelling the *producing* BUILD file uses (`build/out`) as well as the emitted one (`app/build/out`).

Mistakes are hard errors that name what was available, never silent no-ops: a transform matching nothing, a dict key matching nothing, one destination with several candidates, and two files landing on one path.

## Also here

- **Inline dep filters accept globs.** Five call sites open-coded `Path::new(f) == rel` (exact paths only); they now share one `PathFilter`, which falls back to an exact match for any pattern that will not compile — so every filter that worked before selects the same files. The `[a,b]` split is brace-aware so glob alternation (`{a,b}`) is not torn in half.
- **`heph inspect outputs <addr>`** prints the paths a target produces. Diff a group against its deps to answer "why did that file land there?".
- **`example/group/BUILD`** demonstrates both forms. Its existing `deps = ["t1"]` was never a valid ref (pre-existing breakage) and is fixed to `":t1"`.

## Testing

`lint` clean. Full `tst`: **94 result lines, 2536 passed, 0 failed.**

New coverage — 38 core (transform/view/filter), 25 group driver, 18 e2e. The e2e set proves what unit tests cannot: a file produced at `build/out/server` is readable by a consumer at `ws/lib/server`; changing `prefix` moves the consumer's `hashin`; an unchanged transform keeps it stable; the group stays uncacheable; and transitive deps still reach consumers once the group is concrete (`collect_transitive_deps` branches on driver name, not `transparent` — easy to break by "tidying").

Verified against the real binary:

```
$ heph inspect outputs //group:dist      $ heph inspect outputs //group:server
release/lib.txt                          bin/server.txt
release/renamed.txt
```

## Reviewer notes

- The review board (`product-vision`, `hermeticity`, `compatibility`, `code-quality`) was **not** consulted — the authoring session disallowed subagents. `compatibility` is the one worth a look: this adds the `inspect outputs` subcommand (additive) and makes `output_artifact_to_pb` fallible (`Content::View` holds a live `Arc<dyn Content>` into the host cache and has no wire form; it is unconstructible from a plugin, and this reports rather than panics because the caller is a cdylib where a panic aborts).
- A relocating group used as a `tools` entry fails `link.rs`'s "exactly 1 FilePath" check with a confusing message, since its declared paths are globs. Left as-is, out of scope.
- Symlink *targets* are not rewritten. Relative symlinks interior to a tree survive `strip_prefix`/`prefix` (both shift every path equally) but can dangle if `include`/`exclude`/`rename` move only one end. Documented in the module header.

https://claude.ai/code/session_014LftdsamVrvndpbUZ9WUjK

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -867,6 +867,7 @@ dependencies = [
  "tempfile",
  "tokio",
  "tracing",
+ "wax",
  "xxhash-rust",
 ]
 
--- a/crates/builtins/src/plugingroup/mod.rs
+++ b/crates/builtins/src/plugingroup/mod.rs
@@ -677,6 +677,7 @@ mod tests {
                     kind: hcore::hartifactcontent::WalkEntryKind::File {
                         data: Box::new(std::io::Cursor::new(Vec::new())),
                         x: false,
+                        size: 0,
                     },
                 })
             })))
--- a/crates/builtins/src/plugingroup/mod.rs
+++ b/crates/builtins/src/plugingroup/mod.rs
@@ -1,11 +1,47 @@
+//! The `group` driver: re-export other targets' outputs, optionally filtered
+//! and relocated.
+//!
+//! A group has two modes, and which one it is in follows from its config alone:
+//!
+//! * **Aggregate** (no transform configured) — the group is `transparent`, and
+//!   the engine splices its members straight into each consumer's input list.
+//!   It never executes and never produces an artifact of its own. This is the
+//!   original behaviour and stays byte-for-byte what it was.
+//!
+//! * **Filter/relocate** (any of `include`/`exclude`/`strip_prefix`/`prefix`/
+//!   `rename` set) — the group becomes a real target that produces
+//!   [`ContentView`] artifacts: zero-copy windows onto its deps' artifacts with
+//!   rewritten paths. It is marked uncacheable, so no revision, no blob, and no
+//!   remote push is ever created for it; the deps stay cached exactly as
+//!   before, and the group is re-derived from a path list on each build.
+//!
+//! The alternative — an exec target running `cp` — costs a sandbox, a
+//! subprocess, and a second full copy of every byte in both the local and the
+//! remote cache. This costs a few string operations.
+//!
+//! It cannot stay `transparent` in the second mode. A transparent target is
+//! expanded away *before* `hashin` is computed, so its config never reaches a
+//! consumer's cache key — membership changes are caught only because they
+//! change which dep hashouts get folded in. A path transform changes no dep's
+//! hashout, so a transparent one would be invisible to the cache and a consumer
+//! would happily reuse an entry built against the pre-relocation layout. Being
+//! a real target is what puts the transform in the consumer's key, via the
+//! group's own def hash and its artifacts' derived hashouts.
+
 use anyhow::Context as _;
 use async_trait::async_trait;
+use hcore::hartifactcontent::{
+    Content as _, PathTransform, Rename, RenamePlan, SourcePaths, ViewContent,
+};
 use hcore::hasync::Cancellable;
 use hplugin::driver::TargetAddr;
 use hplugin::driver::{
     ApplyTransitiveRequest, ApplyTransitiveResponse, ConfigRequest, ConfigResponse, ParseRequest,
-    ParseResponse, RunRequest, RunResponse,
-    targetdef::{CacheConfig, Input, InputMode, TargetDef},
+    ParseResponse, RunRequest, RunResponse, outputartifact,
+    targetdef::{
+        CacheConfig, Input, InputMode, Output, TargetDef,
+        path::{CodegenMode, Content as PathContent, Path},
+    },
 };
 use hplugin::htspec::Spec;
 use std::sync::Arc;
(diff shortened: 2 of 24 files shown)


raphaelvigee:
## Performance follow-up (aa3eb915)

Two defects found while measuring, both fixed in the pushed commit.

### Blocker: `ViewContent::reader()` buffered the whole archive

It packed the rewritten tree into a `Vec` and returned a `Cursor` — an unbounded allocation on the one path whose purpose is to avoid copying. Now packs on a writer thread into a `std::io::pipe`: **O(1) memory** (one pipe buffer + a 512-byte header) regardless of artifact size, with pipe backpressure throttling a slow consumer. Packer failures surface as read errors (joined at EOF), not a truncated archive.

`tar::Builder` still packs, so GNU long names/checksums/symlinks stay correct — but it needs each entry's size before its bytes, which is *why* the buffering was there. `WalkEntryKind::File` now carries `size`; all 8 producers already knew it (tar from the header just parsed, files from `stat`, in-memory from `len`).

### Regression: `PathFilter` compiled every pattern as a glob

`wax::Glob::new` builds a regex NFA (~19µs), and `plugin-go` emits one exact-path filter **per Go source file** — thousands per build. Measured on that shape (2000 inputs × 200 paths):

| | |
|---|---|
| old exact-compare (baseline) | 7.72 ms |
| all-glob compile + match (as first pushed) | **66.99 ms — 8.1× regression** |
| literal fast path (now) | **6.34 ms** |
| unfiltered inputs | 0.42 ms / 400k calls |

A pattern with no glob metacharacter is now kept as a literal and string-compared, never compiled. `matches` also takes `to_string_lossy` (borrowed) instead of building a slash-normalised `String` per path — which is why it lands *below* the baseline it replaced.

### Also: memoized the path mapping

Resolving reads the source's path list — a header-only scan of the whole archive for a tar-backed artifact — and consumers ask for both `entry_paths()` and `walk()` (collision detection, then materialization), so it was paid twice. Artifacts are immutable for the life of a build.

### Net performance picture

- **Disk** — a relocating group writes zero cache bytes. The `cp`-target alternative writes a full second copy locally *and* pushes it remotely: `2N + network` → `0`.
- **Memory** — reader O(artifact) → O(1); no per-path allocation in filtering. Residual: the mapping is O(entries), ~11k `PathBuf` pairs for the Go SDK, now computed once.
- **CPU** — aggregate groups bit-for-bit unchanged; filtering ~18% faster than before this PR; a relocating group costs one header-only scan plus a path rewrite versus a sandbox, fork/exec, byte copy, tar pack and cache write.

New tests pin all three: a 256 MiB view read 4 KiB at a time, packer failure propagation, and the source path list being scanned exactly once.

**Not measured:** `heph-bench`. Per CLAUDE.md it belongs in CI, where a baseline-vs-candidate comparison beats a local run's noise floor. The above are targeted micro-benchmarks of the touched paths, not a whole-build number. If the perf job flags anything, mapping resolution on the materialization path is where I'd look first.

`lint` clean; full `tst`: **2540 passed, 0 failed**.

raphaelvigee:
## Perf verdict (93977bee)

**No regression on the current head.** Three defects were found and fixed during review; here is where every touched path now stands.

| Path | Status |
|---|---|
| Inline dep filters, exact path (what `plugin-go` emits) | was **8.1x slower**; now **6.34 ms vs 7.72 ms baseline** — ~18% faster than before the PR |
| Unfiltered inputs (common case) | unchanged — 0.42 ms / 400k calls |
| `ViewContent::reader()` memory | was **O(artifact)**; now **O(1)** |
| `PassthroughContent::walk` (every source file) | was **+7%**; now at parity |
| Aggregate `group` (no transform) | bit-for-bit unchanged, def hash pinned identical by test |
| Disk | strictly better — a relocating group writes zero cache bytes vs a full duplicate, local **and** remote |

### The one this commit fixes

Adding `size` to `WalkEntryKind::File` put a `stat` into four `walk` implementations, and two of them resolved the same path twice — `stat` by path, then `open` of the same file. One is hot: `PassthroughContent::walk` is the `@heph/fs:file` route, once per source file per materialization. Over 20k files:

    open only            (before this branch)  227 ms
    stat(path) + open    (as first written)    245 ms   <- +7%
    open + fstat(handle) (now)                 226 ms

`fstat` on a descriptor the function already holds is free next to the `open` it rides on. Same fix in `OutputArtifact::walk`'s `Content::File` arm. `serve.rs` already read from the open handle; `FileContent` is the process-log diagnostic path and is untouched.

### Not measured

`heph-bench` — no whole-build number from me. Per CLAUDE.md it belongs in CI, where baseline-vs-candidate beats a local run's noise floor. Everything above is a targeted micro-benchmark of a specific path. If the perf job flags something, mapping resolution on the materialization path is the first place to look.

`lint` clean; full `tst`: **2540 passed, 0 failed**.

raphaelvigee:
## heph-bench Tier A: no regression

Ran locally since CLAUDE.md permits it when explicitly asked. Baseline is a worktree at `origin/master` (`fcec1f23`) — the true N-1 for this PR, not a fetched release. Candidate is `93977bee`. One corpus (1500 targets / 100 packages / 8 layers / fan-out 4, seed 0), shared by both sides, candidate and baseline alternating rep-by-rep on the same seed. Warmup 1, 7 reps (CI's count). `compare` at CI defaults: 8% threshold, `noise_k` 2.

| Scenario | Baseline (ms) | Candidate (ms) | Delta | Noise band | Effective threshold | Verdict |
|---|---:|---:|---:|---:|---|
| cold | 6338.5 (n=7) | 6245.3 (n=7) | **-1.5%** | 3.1% | 8.0% | ok |
| full-hit | 205.9 (n=7) | 207.8 (n=7) | **+0.9%** | 10.7% | 10.7% | ok |
| incremental | 204.6 (n=7) | 210.3 (n=7) | **+2.8%** | 11.6% | 11.6% | ok |

`compare` exits 0.

### How much to read into it

Every delta sits inside its own noise band, so the honest reading is **no measurable difference** — not "2.8% slower on incremental" and not "1.5% faster on cold".

`full-hit` and `incremental` came back with noise bands of 10.7% and 11.6%: this machine's variance swamps anything under ~10% there, so those two would only have caught a gross regression. **`cold` is the informative row** — a 3.1% band, and the candidate landed slightly under baseline. That is the scenario that exercises the change I was least sure about, since every source file is materialized through `PassthroughContent::walk` and its new `fstat`. It does not show up.

### What this run does NOT cover

- **`ViewContent` / relocating groups** — the synthetic corpus has none. Zero coverage; the streaming reader and mapping memoization are backed only by unit tests plus the 256 MiB streaming test.
- **Glob dep filters** — no `[...]` deps in the corpus. The 8.1x regression fixed in aa3eb915 **would not have been caught by this bench at all**; that took a targeted micro-benchmark on `plugin-go`'s shape.
- **Tier B** — not run. Needs prebuilt dists plus the Go plugin cdylib on both sides, and it silently measures nothing without a declared Go variant.
- **Scale** — 1500 targets vs CI's 4000, on a machine that has been compiling Rust all session.

CI's `perfbench` job on quiet runners remains the authoritative check. This says the engine-wide cost of the PR is not distinguishable from noise, with two of three scenarios too noisy locally to say much either way.

PR #407, opened 2026-08-22
feat(scratch): reference a scratch cache by addr from a consuming target

raphaelvigee:
Second of four (stack #408), on top of #403. That one added the **declaration**;
this one adds the **reference** — `scratch = ["//build:gocache"]` on a consuming
target — and every check that does not need storage. Mounting is #412.

Nothing observable happens yet beyond `heph query`. That is deliberate: the
validation lands before anything can write.

## The reference is `hashed: false, runtime: false`

The one combination nothing else in the tree uses, and both halves are
load-bearing:

- **`runtime: false`** — it materializes no artifacts, because a declaration has
  none. `link.rs:36` filters these out of `LinkedTargetDef`, so a scratch never
  reaches the sandbox as an input.
- **`hashed: false`** — it must not touch the consumer's cache key, because a
  target's outputs are required to be identical whether its scratch is warm,
  cold, or absent.

That second half is the property most likely to be "fixed" into a bug later, so
it is asserted three ways: the def hash is byte-identical with and without a
reference; referencing a scratch does not change what a target produces; and
bumping a scratch's `version` leaves every consumer's cached result a hit.

The tempting alternative is to fold the declaration into the consumer's `hashin`,
so bumping `version` rebuilds everything using the cache. That is an over-hash. If
outputs really are identical warm-or-cold, a fresh slot changes nothing and the
rebuild is pure waste; if they are *not*, the target is already broken and
rebuilding is not the fix. So a `version` bump gives every consumer a fresh empty
slot and invalidates nothing — which is exactly what you want when the reason for
bumping it is "the old cache had gone bad".

The edge still exists in the graph, which buys `heph query revdeps //build:gocache`
— "who shares this cache?", the question every serialization surprise starts with
— and turns a bad addr into an ordinary `TargetNotFoundError` rather than a
bespoke "unknown scratch name".

## Compatibility: nothing new crosses the plugin ABI

No proto message, no `TargetDef` field, no `ABI_SEMVER` bump. The reference rides
on `Input.annotations` — host-visible, and already the producer→host channel
(`READ_ONLY_ANNOTATION` and `STAGE_PER_FILE_ANNOTATION` are the precedent) — and
the settings are read from the declaration's spec config rather than travelling on
the edge. A third-party driver participates without being recompiled.

The cost is stringly-typed settings on the wire, which is the trade `read_only`
already made. `driver-support/src/scratch.rs` holds the key constants so they have
one definition.

*(Mounting does need ABI surface — see #412. Declaration and reference do not.)*

## Validation is in the engine, not the driver

A driver sees its own target's config and nothing else. The properties that matter
are only visible across a *resolved set* of references, which is a host operation —
and putting the checks there means every driver gets them, not just pluginexec:

- a referenced addr that is not a `scratch` target — names both ends, and suggests
  `deps`, since that is the likely intent
- two references claiming one environment variable, which would silently make one
  of two real caches unreachable. The fix is on a *declaration*, so the error says
  so; there is nothing to change at the use site
- two mounts overlapping, which would write one cache through the other. Compared
  component-wise, not by string prefix, so `.cache/go` and `.cache/golang` are
  correctly **not** a collision — a `starts_with` would emit an error the author
  cannot act on
- the same scratch referenced twice, rather than quietly collapsing it

## Tests

22, across pluginexec parse, the engine's resolver, and e2e. The def-hash ones are
the load-bearing set — note they compare the **def hash** and not `hashout`: a
target whose key moved still produces identical bytes, so a hashout comparison
would pass while the cache missed on every run. There is a precondition asserting
the comparison discriminates at all, so the equality checks cannot pass vacuously.

---

### DIFF
--- a/crates/builtins/src/lib.rs
+++ b/crates/builtins/src/lib.rs
@@ -4,6 +4,7 @@
 //!
 //! - `pluginfs` — the `fs` provider + driver (filesystem targets).
 //! - `plugingroup` — the `group` driver (aggregate targets).
+//! - `pluginscratch` — the `scratch` driver (persistent cache-directory declarations).
 //! - `pluginstatictarget` — in-memory static target provider (tests/wiring).
 //! - `plugintextfile` — the `textfile` driver.
 //! - `pluginhostbin` — the host-binary provider + driver.
@@ -11,6 +12,7 @@
 pub mod pluginfs;
 pub mod plugingroup;
 pub mod pluginhostbin;
+pub mod pluginscratch;
 pub mod pluginstatictarget;
 pub mod plugintextfile;
 
--- a/crates/builtins/src/pluginscratch/mod.rs
+++ b/crates/builtins/src/pluginscratch/mod.rs
@@ -1,0 +1,645 @@
+//! The `scratch` driver: declares a named, mutable, non-hermetic cache directory
+//! that persists across runs and is shared by every target referencing it.
+//!
+//! A scratch directory is state carried between runs to make a build *faster*,
+//! under one invariant that everything else here follows from:
+//!
+//! > A target must produce identical outputs whether its scratch directories are
+//! > warm, cold, or absent. Losing one is always a slowdown and never a wrong
+//! > answer.
+//!
+//! This driver is only the **declaration**. It has no inputs, no outputs, never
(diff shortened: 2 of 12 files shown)


raphaelvigee:

PR #422, opened 2026-08-24
raphaelvigee/exec runners host exec

raphaelvigee:


raphaelvigee:
Folded into #418, where the behaviour it replaced was introduced — the stack is unmerged, so the refusal it deleted should simply never exist rather than be added and removed across two PRs.

The `ExecSession::stream` half is dropped entirely: with exec across the seam being batch, its only caller was its own test.

PR #426, opened 2026-08-26
fix(go): copy with whatever `cp` the environment has

raphaelvigee:
The thirdparty download and the std install both picked `cp` flags from `uname`, and `-X` (suppress xattr copy) is BSD-only — GNU `cp` rejects it outright. A Mac inside a `devenv` breaks: `uname` says Darwin while `PATH` leads with nixpkgs coreutils.

Not a corner case. `PATH` is the host's under `gotool = "host"` and entirely the runner's under a `runner:`, so nothing in the sandbox can infer the dialect from the OS.

`target_std` had already hit this and answered with `/bin/cp` — trading the bug for a worse one: a host absolute path, in a sandbox, that a container image or NixOS box need not have. The download target copied the `uname` branch without the workaround.

Both are now a plain, PATH-resolved `cp -R` / `cp -RL`.

### The trade

`-X` was there for a reason: BSD `cp` copies xattrs (`com.apple.quarantine` on Go's mod cache files) and that EPERMs on a FUSE overlay rejecting xattr writes. Dropping it exposes **macOS under `fuse.enabled` only** — elsewhere the copied xattrs are inert metadata on scratch files, and Linux has none to copy. FUSE is already an untested gap here.

### Why not the obvious repairs

Both fail, and it's worth recording:

- **A shared `cp` flag** — there isn't one. GNU `cp` on APFS reflinks by default and `clonefile()` carries xattrs regardless of `--no-preserve=xattr`; only `--reflink=never` suppresses it, which BSD `cp` rejects.
- **`tar --no-xattrs`** — bsdtar accepts it but still carries mac metadata. The flag that works (`--no-mac-metadata`) is rejected by GNU tar.

### Tests

One per site, pinning the copy as a plain `cp` with no `/bin/cp` and no `uname`, so a platform branch cannot quietly return. 483 pass in `plugin-go`; lint clean.

https://claude.ai/code/session_01VaazTTm6xTUQxxpxhY9tJW

### DIFF
--- a/crates/plugin-go/src/plugingo/target_std.rs
+++ b/crates/plugin-go/src/plugingo/target_std.rs
@@ -73,17 +73,25 @@ pub fn install_spec(
         // populate pkg/. `-L` dereferences symlinks: staged dep files may be
         // symlinks into the read-only artifact cache, and a plain copy would
         // leave LGOROOT pointing back at it (so `chmod -R u+w` would try to
-        // mutate the cache). BSD/macOS `cp` also copies xattrs and trips EPERM
-        // on the FUSE overlay; `cp -X` suppresses that (GNU `cp -X` differs, so
-        // branch on uname). Use the absolute system `cp` (`/bin/cp`) so the flag
-        // set matches the platform regardless of `$PATH` — under the host
-        // toolchain (`gotool = "host"`) the sandbox inherits the host `PATH`,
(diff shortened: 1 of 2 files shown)


raphaelvigee:
Folding this into #425 instead — it lands there as 70608bd5.

PR #432, opened 2026-08-29
feat(scratch): make the store inspectable and reclaimable

raphaelvigee:
Fifth of the scratch stack (#408), on top of #431. The first four made scratch
caches work; **nothing bounded them**. A scratch is keyed by a declaration rather
than an input hash, so there is no `hashin` to age out and no `cache.history` to
trim against — it grows until something removes it, and until now nothing did.

```
heph tool scratch ls          # what is there, largest first
heph tool scratch path ADDR   # where it is
heph tool scratch rm ADDR     # drop one; --all drops every one
heph tool clean --scratch      # same, for someone already in `clean`
heph tool gc --scratch-max-size 50GiB --scratch-max-age-days 30
```

## The sweep lives in `heph tool gc`

Not behind a flag nobody sets. A GC that reclaimed target revisions and left the
one part of the store that grows without limit would be misleading about what it
does.

**Whole caches, never partial trims.** heph cannot know which of a foreign tool's
entries are hot, and guessing would quietly degrade a cache while claiming to
manage it. Dropping is honest and self-correcting: the next build repopulates what
it actually needs. And removing a scratch is *always* safe — a target's outputs are
identical whether its scratch is warm, cold or absent, so it costs time and
nothing else.

## The store describes itself

Each slot records what it came from in a small `slot.meta`. Without it a listing
could only print hashes, or would have to resolve the whole graph to map them back.
100 bytes per slot buys a store that stays listable and removable after the target
that made it is deleted or renamed — the same property `heph tool clean` already
has for addr-only selections, and for the same reason.

Two consequences worth checking:

- A slot whose meta is unreadable is **still listed** (it is occupying disk) and
  **still removable by `--all`**, but is never matched by name — it cannot be
  named, so naming must not silently miss it and clearing everything must not
  silently leave it.
- Sizes are measured **per scope**, so they mean the cache's own footprint and not
  that plus heph's bookkeeping. A size that silently included the meta would make
  every eviction threshold slightly wrong in a way nobody could see. (This came out
  of a test: the sweep evicted two slots where one should have sufficed.)

## `--scratch=off`, and why it forces a rebuild

`off` runs every target with its caches absent — the audit mode for the contract
the whole feature rests on. The *whole path* is skipped, not merely
mounted-and-empty: no spec resolution, no lock, no directory. That is what makes
it a faithful "as if absent".

**It implies `--force`, and has to.** Scratch deliberately never reaches `hashin`,
so without a rebuild an off-run is a plain cache hit that replays the result built
*with* a warm cache — the audit would pass by reading exactly the answer it is
supposed to re-derive. I found this because the first version of the test failed:
it asserted `absent` and got `mounted`.

The vacuous case is now asserted directly, before the forced one, so the
implication cannot be quietly simplified away later.

This changes no cache key: the audit run writes its revision under the same one, so
a divergence surfaces as a rebuilt-hashout mismatch rather than as two unrelated
entries.

`--scratch` is a build flag next to `--force` rather than a `heph tool scratch`
subcommand, because it modifies a build rather than being one.

## Compatibility

Additive. New subcommand, two new `gc` flags, one new `clean` flag, one new global
flag defaulting to today's behaviour. No proto change, no ABI bump, no change to
any existing cache entry or on-disk layout.

## Tests

11 store unit tests, 7 CLI unit tests, 2 e2e. The e2e pair covers the audit mode
(including the vacuity trap) and a build creating a slot that is then listed and
removed by name with no BUILD files read.

---

### DIFF
--- a/crates/builtins/src/lib.rs
+++ b/crates/builtins/src/lib.rs
@@ -4,6 +4,7 @@
 //!
 //! - `pluginfs` — the `fs` provider + driver (filesystem targets).
 //! - `plugingroup` — the `group` driver (aggregate targets).
+//! - `pluginscratch` — the `scratch` driver (persistent cache-directory declarations).
 //! - `pluginstatictarget` — in-memory static target provider (tests/wiring).
 //! - `plugintextfile` — the `textfile` driver.
 //! - `pluginhostbin` — the host-binary provider + driver.
@@ -11,6 +12,7 @@
 pub mod pluginfs;
 pub mod plugingroup;
 pub mod pluginhostbin;
+pub mod pluginscratch;
 pub mod pluginstatictarget;
 pub mod plugintextfile;
 
--- a/crates/builtins/src/pluginfs/mod.rs
+++ b/crates/builtins/src/pluginfs/mod.rs
@@ -1533,6 +1533,7 @@ mod tests {
             stdout: None,
             stderr: None,
             sandbox_dir: root,
+            scratch: vec![],
         }
     }
 
--- a/crates/builtins/src/plugingroup/mod.rs
+++ b/crates/builtins/src/plugingroup/mod.rs
@@ -727,6 +727,7 @@ mod tests {
                     stdout: None,
                     stderr: None,
                     sandbox_dir: Default::default(),
+                    scratch: vec![],
                 },
                 &ctoken(),
             )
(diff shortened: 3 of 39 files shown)


raphaelvigee:

PR #443, opened 2026-08-29
feat(coreutils): add sed

raphaelvigee:
-

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1067,6 +1067,7 @@ dependencies = [
  "grep-printer",
  "grep-regex",
  "grep-searcher",
+ "regex",
  "regex-syntax",
  "tar",
  "tempfile",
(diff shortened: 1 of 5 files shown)


raphaelvigee:
Superseded by consolidation. The applet halves of this PR (#442 tar/gzip/zstd, #443 sed, #444's `tmpl`) now land in two PRs instead of four:

- **#453** — `grep`, `find`, `xargs`, `sed`, `tar`, `gzip`, `zstd`. They touched the same four files and asked the same review question three times.
- **#440** — the `template` rule and the `tmpl` applet together, since they are one feature: a shared renderer used two ways.

No content was dropped — the restacked tree is byte-identical to the eight-commit version, and every layer builds, lints and passes its unit tests on its own. Fewer layers also matters now that stacked PRs get no CI unless labelled (#449).

PR #450, opened 2026-09-02
ci: add Codecov integration via grcov

raphaelvigee:
Adds line coverage to CI: measured with [grcov](https://github.com/mozilla/grcov) over rustc's own `-Cinstrument-coverage`, uploaded to Codecov from a `linux/amd64` and a `darwin/arm64` leg.

## Why grcov, of the three

| | |
|---|---|
| **kcov** | DWARF breakpoints on line entries. Linux-only, so it cannot run on `aarch64-apple-darwin` at all, and its accuracy degrades exactly where this workspace lives — monomorphised generics, inlining, async state machines. |
| **cargo-tarpaulin** | Default engine is ptrace, so also Linux-only. ptrace is also the worst-placed mechanism for a program whose interesting seams are spawned subprocesses and `dlopen`ed cdylibs. |
| **grcov** ✅ | Consumes the `.profraw` rustc itself emits. Nothing platform-specific, so it runs on all three supported targets; it counts regions rather than line breakpoints; and because the counters live in the instrumented binary rather than in a tracer, an instrumented child process or cdylib writes its own profile and is counted with everything else. `llvm-tools` was already a pinned toolchain component and carries the `llvm-profdata` it needs. |

## What's here

- **`cov`** (devenv.nix) — one entrypoint, identical locally and in CI, same shape as `e2e`. Delegates the suite to `tst` rather than restating its package list, so the report always describes the suite that actually gates this repo. `cov -p engine` for a narrow local loop.
- **`coverage` job** — matrix over `linux/amd64` + `darwin/arm64`, each uploading under its own Codecov flag; Codecov merges them.
- **`codecov.yml`** — validated by Codecov's own validator.
- **`scripts/coverage-report.py`** — filters, rewrites the lcov, and reports: a worst-covered-first per-crate table, and `summary.json` (sorted keys, no timestamps) as the agent surface.

## `#[cfg(test)]` modules are stripped — the one thing to push back on if you disagree

**39% of this tree's Rust lines sit inside a `#[cfg(test)]` module**, and source-based coverage instruments them along with the code they test. Neither Codecov's `ignore:` nor grcov's `--ignore` can reach them: both match on paths, and these live inside production files.

Left in, they do more than inflate the headline — they invert the signal. `.claude/testing.md` requires every change to ship with a test, so *every* PR here is production-code-plus-test-module, and patch coverage (the number a reviewer actually reads) would rise in proportion to how much test code the PR added.

Measured on `crates/core`: **86.37% with them, 73.22% without** — 2,989 of its 5,614 instrumented lines were test code, all counted as covered.

Done now rather than later because Codecov freezes the project baseline on first upload.

The filtering is computed from the source rather than with grcov's `--excl-start`, because the obvious regex spelling (`^#\[cfg\(test\)\]$` … `^\}$`) is wrong on this tree in a direction that fails silently: 14 of the 198 column-0 `#[cfg(test)]` sites are not module blocks (`mod foo;` declarations, a `use`, two `const`s), and for those the range would run to whatever the next column-0 `}` happened to be, **deleting production lines from the denominator**.

## Coverage gates nothing, and that is the point

Every Codecov status is `informational` and the job is absent from `release`'s `needs:`. A percentage that blocks a merge buys tests that move the percentage.

The corollary is that nothing downstream can tell a broken measurement from a real drop — a report that measured almost nothing would publish as a plausible number and read as a regression somebody caused. So `cov` exits non-zero on every way it knows of measuring nothing:

- no profiles written
- **a zero-length profile** — `llvm-profdata` skips these in complete silence (measured: exit 0, no stderr), so without this check ENOSPC or an OOM kill mid-write yields a smaller, entirely plausible number
- a profile written outside `$LLVM_PROFILE_FILE`
- a profile grcov skipped (its stderr is scanned; the keyword list includes `truncated`, which is llvm-profdata's actual wording for a partial file and matches none of the more obvious words)
- zero tests run
- a report under the size floors, or the named canary file unhit
- no `llvm-profdata` where the pinned toolchain should have one

## Two platform legs

Per the portability rule in CLAUDE.md, this was put to @raphaelvigee rather than settled by the implementation. Coverage is per-line, so a Linux-only report leaves `crates/proc/src/proc_exec/imp_macos.rs` — ~790 lines of the most delicate code in the tree — *absent* from the report rather than uncovered in it, which Codecov reports as "nothing to cover": a green patch check over entirely unmeasured code. The arch axis is deliberately not covered; x86_64-vs-aarch64 differences here are memory ordering and codegen, which line coverage cannot see.

## Three traps worth naming

All three bit during development and all three are now gated by a test:

1. **An env `RUSTFLAGS` *replaces* `.cargo/config.toml`'s `build.rustflags`** rather than merging — cargo takes its extra flags from exactly one source. That drops `-Cforce-frame-pointers=yes` loudly (it reddens `pprof_dump`'s own test) and `--cfg tokio_unstable` *silently*. `cov` reads the configured list rather than hardcoding a copy, which would be the same bug one refactor later.
2. **Without an explicit target, cargo instruments build scripts and proc-macros too** — so rustc dlopens an instrumented `htspec-derive` and every compile in the workspace drops a `.profraw`, burying the test profiles and reporting build-script lines as covered code.
3. **`find … | head -n 5` inside a command substitution takes SIGPIPE**, and `pipefail` + `set -e` turn that into a bare exit 141 with no message — in the one scenario the check exists to report.

## Tests

- `tests/coverage_report.rs` (13) drives the real report script: floors, canary, the lcov parse (duplicate `SF:` accumulation, checksum-bearing `DA:`), and **both directions** of the `#[cfg(test)]` stripping — that test modules go and that production code stays.
- `tests/coverage_gate.rs` (12) guards the configuration. Each `cov` guard is asserted as *branch-plus-`exit`* rather than as a keyword: the first version of that test asserted vocabulary and passed over a `cov` with every check deleted, because names like `profraw_count` survive in the lines that report them.
- Parsers shared by `lint_gate` and `coverage_gate` move to `tests/common/mod.rs`.

Every gate assertion was mutation-tested — each goes red when its invariant is broken and green when restored, including "guards present but exits unhooked".

## Verification

`lint` exits 0. All 34 tests pass. `cov -p core --lib` ran end to end on darwin/arm64 (grcov 0.9.1 reads rustc 1.96 profiles cleanly). The stray-profile guard was verified end to end by planting one; the SIGPIPE fix and the `-empty` predicate were verified in isolation. The workflow YAML parses and `codecov.yml` passes Codecov's validator.

**What CI will tell us that this cannot:** run #1 is a full cold instrumented compile (`-Cinstrument-coverage` shares no kache key with the normal build), so watch the darwin leg's wall time against the 90-minute cap, and the `du -sh`/`df -h` lines `cov` now prints for the real profile volume on the disk-constrained Linux runner.

**Needs `CODECOV_TOKEN`** in repo secrets before the upload will do anything; without it the legs still measure and write the report to the step summary, and log a warning.

https://claude.ai/code/session_011MThcHarSHAYunrefuzcPw


raphaelvigee:
## Re: "isn't this reinventing the wheel?"

Fair challenge, so I tried the wheel rather than defending the code. `cargo-llvm-cov` is the obvious fourth option — I'd treated the three named tools as the whole option space, which was mine to surface rather than file away.

**On paper it replaces most of `cov`.** Verified, not assumed — `cargo llvm-cov show-env` emits:

```
RUSTFLAGS='--cfg tokio_unstable -Cforce-frame-pointers=yes -C instrument-coverage --cfg=coverage'
```

That is the whole `build.rustflags` composition problem, solved. Plus host/target splitting, profile management and cleanup, and — the one that actually matters — reading profiles with the toolchain's *own* `llvm-profdata`, which makes the grcov-vs-rustc version skew I documented as a risk structurally impossible.

So I rewrote `cov` on top of it. **Then reverted it**, because it does not survive contact with kache:

- kache is this repo's `RUSTC_WRAPPER`; cargo-llvm-cov installs its own by default. `--no-rustc-wrapper` keeps kache — but that wrapper is also how it normally avoids instrumenting build scripts and proc-macros, so `--target` has to cover that instead.
- `cargo llvm-cov report` then resolves objects and sources by **the path the compiler recorded**, and kache normalises those to `/kache/…`.

Measured on `-p core --lib`, same tree, same profiles:

| | result |
|---|---|
| grcov | 2,625 lines across `crates/core`, 73.26% |
| `cargo llvm-cov report` | **two registry files, exit 0** |

A report of almost nothing, no error. Exactly the silent-green class this whole PR is built against.

grcov is immune because it is handed `--binary-path` and `--source-dir` explicitly and relativises with `--ignore-not-existing`, so it never depends on a recorded absolute path matching this machine. That is what the plumbing buys, and it's now written down in `devenv.nix` next to the code — with a note to revisit if kache gains a no-remap mode or cargo-llvm-cov gains explicit binary/source paths.

## And `#[coverage(off)]`?

Also checked rather than assumed — still unstable on this repo's toolchain:

```
error[E0658]: the `#[coverage]` attribute is an experimental feature
 --> t.rs:1:1
  = note: see issue #84605 for more information
rustc 1.96.0 (ac68faa20 2026-05-25)
```

Adopting it needs a nightly toolchain for the coverage legs, `#![feature(coverage_attribute)]` in 42 crate roots, and an annotation on all 184 `#[cfg(test)] mod tests` blocks — which a new module can silently forget, i.e. the same failure the filter exists to close, moved somewhere with 184 chances to happen. When it stabilises the attribute becomes the better answer (regions excluded at instrumentation time rather than after the fact) and `scripts/coverage-report.py` can go. Noted in the file.

No behaviour change in cafc72b9 — comments only.

